PR #3, opened 2019-04-09
Resolve container versions and OpenAPI spec links through host-meta.json (RFC-6415)

schrieveslaach:
When requesting the list of apps, PREvant tries to resolve .well-known/host-meta.json for each service which should provide versioning information and the link to the OpenAPI specification.

### DIFF
--- a/api/src/models/service.rs
+++ b/api/src/models/service.rs
@@ -24,10 +24,13 @@
  * =========================LICENSE_END==================================
  */
 
+use crate::models::web_host_meta::WebHostMeta;
+use chrono::{DateTime, Utc};
 use regex::Regex;
 use serde::ser::{Serialize, Serializer};
 use serde::{de, Deserialize, Deserializer};
 use std::collections::BTreeMap;
+use std::net::IpAddr;
 use std::str::FromStr;
 use url::Url;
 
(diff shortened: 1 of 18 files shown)


api/src/models/service.rs
```diff
@@ -36,8 +39,25 @@ pub struct Service {
...
+#[derive(Clone, Debug)]
+struct ServiceEndpoint {
+    internal_addr: IpAddr,
+    exposed_port: u16,
+}
+
+impl ServiceEndpoint {
+    fn to_url(&self) -> Url {
+        Url::parse(&format!(
+            "http://{}:{}/",
```
x1B: I guess SSL to the containers is not supported anyway? Not that it matters much when using virtual networks.

api/src/services/apps_service.rs
```diff
@@ -58,10 +61,70 @@ impl AppsService {
...
+                "Forwarded",
+                format!(
+                    "host={};proto={}",
+                    request_info.host(),
+                    request_info.scheme()
+                ),
+            )
+            .header(
+                "X-Forwarded-Prefix",
+                format!("/{}/{}", app_name, service_name),
```
x1B: One _might_ add recursive Proxy support here (prepending any incoming `X-Forwarded-Prefix`).
Maybe this is premature, but somehow I get the feeling it'll be useful sooner or later.

api/src/models/service.rs
schrieveslaach: Correct, currently PREvant assumes that there is an HTTP endpoint. But, if the container support HTTP redirects, [`reqwest` handles redirects:](https://docs.rs/reqwest/0.9.15/reqwest/#redirect-policy)

> By default, a Client will automatically handle HTTP redirects, detecting loops, and having a maximum redirect chain of 10 hops. To customize this behavior, a RedirectPolicy can used with a ClientBuilder.

In this case PREvant should be able to handle redirects to HTTPS. However, in the future we might have to handle self-signed certificates and stuff like this.

api/src/services/apps_service.rs
schrieveslaach: Do you think of scenarios in which PREvant is running behind another proxy?

schrieveslaach: If so, do you want to add an issue so that we can keep track of it?

x1B: Good call, I opened #4 for this

Issue #5, opened 2019-05-06
Web UI should work in MSIE11


x1B:
There's some JS in the Vue.js UI, preventing MSIE11 users from accessing the PREvant frontend.

Marking as an enhancement, although that's certainly debatable :)

x1B:
Merged with commit c6e0884

PR #6, opened 2019-05-06
(#5) enable MSIE 11 support

x1B:


frontend/webpack.config.js
```diff
@@ -94,6 +95,6 @@ module.exports = {
     performance: {
         hints: false
     },
-    devtool: '#eval-source-map',
+    devtool: '#cheap-source-map',
```
x1B: Cut `app.js` bundle size from ~10m to ~1.8m

schrieveslaach: Very nice. :sunglasses:

PR #7, opened 2019-05-22
Implementing support for secrets

schrieveslaach:
Add configuration to define secrets for specific services for specific apps. These
secrets will be mounted in the container under `/run/secrets/<name>` so that the
service is able to use e.g. database credentials.

### DIFF
--- a/api/src/models/service.rs
+++ b/api/src/models/service.rs
@@ -69,6 +69,7 @@ pub struct ServiceConfig {
     #[serde(deserialize_with = "Image::parse_from_string")]
     image: Image,
     env: Option<Vec<String>>,
+    // TODO: rename this field because it does not match to volumes any more (it is file content, cf. issue #8)
     volumes: Option<BTreeMap<String, String>>,
     #[serde(skip)]
     labels: Option<BTreeMap<String, String>>,
@@ -144,6 +145,16 @@ impl ServiceConfig {
         match &self.labels {
             None => None,
             Some(labels) => Some(&labels),
+        }
+    }
+
+    pub fn add_volume(&mut self, path: String, data: String) {
+        if let Some(ref mut volumes) = self.volumes {
+            volumes.insert(path, data);
+        } else {
+            let mut volumes = BTreeMap::new();
+            volumes.insert(path, data);
+            self.volumes = Some(volumes);
         }
     }
 
--- a/api/src/services/config_service.rs
+++ b/api/src/services/config_service.rs
@@ -25,6 +25,9 @@
  */
 
 use crate::models::service::{ContainerType, Image, ServiceConfig};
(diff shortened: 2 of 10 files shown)


api/src/models/service.rs
```diff
@@ -69,6 +69,7 @@ pub struct ServiceConfig {
     #[serde(deserialize_with = "Image::parse_from_string")]
     image: Image,
     env: Option<Vec<String>>,
+    // TODO: rename this field because it does not match to volumes any more (it is file content)
```
qacwnfq: Are we tracking this TODO anywhere (github issue or JIRA)?

api/src/services/config_service.rs
```diff
@@ -45,6 +52,7 @@ pub struct ContainerConfig {
 pub struct JiraConfig {
     host: String,
     user: String,
+    // TODO: use secstr
```
x1B: I guess that this definitely warrants an issue?

api/src/services/config_service.rs
```diff
@@ -221,14 +298,40 @@ impl TryFrom<&Companion> for ServiceConfig {
...
+        ( $name:expr ) => {{
+            let mut hasher = Sha256::new();
+            hasher.input($name);
+            let img_hash = &format!("sha256:{:x}", hasher.result_reset());
+
+            ServiceConfig::new(String::from($name), Image::from_str(&img_hash).unwrap())
+        }};
+    }
+
+    macro_rules! config_from_str {
```
x1B: I didn't know creating macros would be so simple :+1:

api/src/services/config_service.rs
```diff
@@ -343,22 +446,148 @@ mod tests {
...
     }
+
+    #[test]
+    fn should_return_application_companions_with_specific_app_selector() {
+        let config = config_from_str!(
+            r#"
+            [companions.openid]
+            serviceName = 'openid'
+            type = 'application'
+            image = 'private.example.com/library/opendid:latest'
```
x1B: 🦅 👁 not a big deal, but `opendid`->`openid`?

api/src/services/config_service.rs
schrieveslaach: Rust is just awesome. :wink:

api/src/models/service.rs
schrieveslaach: I opened an issue: #8

api/src/services/config_service.rs
schrieveslaach: That was a leftover... Now, it is fixed.

api/src/services/config_service.rs
schrieveslaach: Nice catch. :+1: :eagle:

PR #9, opened 2019-06-03
Some preliminary support of passing docker run options to services

FabianEmmes:
In my usecase I need to pass parameters (specifically volumes) to `docker run`. This change adds a service option "volumes" to be set in the PREvant configuration. It could be extended to other `docker run` parameters. This PR is meant as a base for discussion of how to add this kind of feature to PREvant.


schrieveslaach:
Closing due to inactivity.

PR #12, opened 2019-07-23
Frontend: Retrieve and Display Logs of Services

schrieveslaach:
This PR adds following features:

- Add an REST resource to retrieve the log messages of a service of an review app with pagination.
- Utilize this resource to display the log messages in the frontend.

### DIFF
--- a/api/res/openapi.yml
+++ b/api/res/openapi.yml
@@ -103,13 +103,7 @@ paths:
     delete:
       summary: Shutdown a review app
       parameters:
-        - in: path
-          name: appName
-          allowEmptyValue: false
-          schema:
-            type: string
-          required: true
-          description: Name of review app to delete
+        - $ref: '#/components/parameters/appName'
       responses:
         '200':
           description: 'List of deleted containers'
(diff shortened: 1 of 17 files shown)


api/res/openapi.yml
```diff
@@ -131,6 +125,42 @@ paths:
...
+        - $ref: '#/components/parameters/appName'
+        - in: path
+          name: serviceName
+          description: Name of the service
+          required: true
+          schema:
+            type: string
+        - in: query
+          name: since
+          description: Date time since when the logs have to retrieved. By default the logs from the beginning are crawled.
```
sofax: Date + time?

api/res/openapi.yml
```diff
@@ -131,6 +125,42 @@ paths:
...
+          schema:
+            type: string
+            format: date-time
+          example: '2019-07-22T08:42:47-00:00'
+        - in: query
+          name: limit
```
FabianEmmes: I was surprised that the limit is a limit on the first `n` lines rather than the last `n` lines. Is it possible to the most current `n` lines also? IMHO this will be a common use case.

sofax: I assume the purpose of the limit parameter is to reduce the memory footprint. But if you want to apply it to the last n lines instead of the first n lines, the whole content of the log file has to be loaded into memory first, before the respective starting line can be determined.

api/res/openapi.yml
```diff
@@ -131,6 +125,42 @@ paths:
...
               schema:
                 $ref: '#/components/schemas/ProblemDetails'
+  /apps/{appName}/logs/{serviceName}/:
+    get:
+      summary: Retrieves the logs from stdout/stderr of the specified container.
+      parameters:
+        - $ref: '#/components/parameters/appName'
+        - in: path
+          name: serviceName
+          description: Name of the service
```
sofax: Shouldn't *appName* be documented here, too?

api/res/openapi.yml
FabianEmmes: You only need a ring-buffer of `n` lines to calculate the last `n` lines. When I wrote the comment, I overlooked the `since`-parameter, which already covers the mentioned use-case, so I don't think a limit for the last `n`-lines is needed anymore.

schrieveslaach: The purpose of the `limit` parameter is to make the REST calls for the frontend simple. The frontend request the first chunk of logs and the REST API responsed with the first chunk and a link header (see [RFC5988](https://tools.ietf.org/html/rfc5988)) to the next page (a.k.a. the next chunk of logs). When the user scrolled to the end of logs, the frontend can use the next link to crawl more logs.

api/res/openapi.yml
schrieveslaach: :+1:

api/res/openapi.yml
schrieveslaach: It is. cf. `$ref: '#/components/parameters/appName'`

api/src/infrastructure/docker.rs
```diff
@@ -586,6 +588,70 @@ impl Infrastructure for DockerInfrastructure {
...
+                };
+
+                let logs = runtime.block_on(
+                    docker
+                        .containers()
+                        .get(&container.id)
+                        .logs(&log_options)
+                        .enumerate()
+                        // Unfortunately, docker API does not support head (cf. https://github.com/moby/moby/issues/13096)
+                        // Until then we have to skip these log messages which is super slow…
```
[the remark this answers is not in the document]
schrieveslaach: I will mention this PR in moby. Maybe the will reopen the issue.

PR #15, opened 2019-10-18
Implement Start and Stop of Services

schrieveslaach:
In order to test the behaviour of services when resources are not available, PREvant now supports the status change of services. Therefore, PREvant offers a new resource which allows to start and stop containers. Additionally, the UI provides buttons, utilizing this resource.

Additionally, this commit cleans up the backend structure and it also uses the store to reflect the change to the application.

### DIFF
--- a/api/res/openapi.yml
+++ b/api/res/openapi.yml
@@ -125,17 +125,40 @@ paths:
             application/problem+json:
               schema:
                 $ref: '#/components/schemas/ProblemDetails'
+  /apps/{appName}/states/{serviceName}/:
+    put:
+      summary: Changes the state of a service
+      parameters:
+        - $ref: '#/components/parameters/appName'
+        - $ref: '#/components/parameters/serviceName'
+      requestBody:
+        description: The new state of the service
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/State'
+      responses:
+        '202':
+          description: The state change has been accepted
+        '404':
+          description: Cannot find app or cannot find service.
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/ProblemDetails'
+        '500':
+          description: Server error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/ProblemDetails'
   /apps/{appName}/logs/{serviceName}/:
     get:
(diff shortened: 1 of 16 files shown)


frontend/src/store/index.js
```diff
@@ -123,6 +123,11 @@ export default new Vuex.Store( {
          } );
       },
 
+      updateServiceStatus( state, { appName, serviceName, serviceStatus } ) {
+         const service = state.apps[appName].find(service => service.name == serviceName);
+         Vue.set( service.state, 'status', serviceStatus );
```
x1B: Should this attribute not be present from the start? What if the service has already be stopped by someone else when loading the UI? But if it _is_ present from the start, `Vue.set` should not be necessary.

api/src/infrastructure/docker.rs
```diff
@@ -641,6 +645,39 @@ impl Infrastructure for DockerInfrastructure {
...
+                let mut runtime = Runtime::new()?;
+                let details = runtime.block_on(c.inspect())?;
+                match status {
+                    ServiceStatus::Running => {
+                        if !details.state.running {
+                            runtime.block_on(c.start())?;
+                        }
+                    }
+                    ServiceStatus::Paused => {
+                        if details.state.running {
```
x1B: Does this mean that updates to the container could be lost, if the Web UI is toggled twice in quick succession, before Docker has implemented the change?

api/res/openapi.yml
```diff
@@ -125,17 +125,40 @@ paths:
...
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/State'
+      responses:
+        '202':
+          description: The state change has been accepted
+        '404':
+          description: Cannot find app
```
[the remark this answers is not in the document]
schrieveslaach: That's correct. I'll fix the description.

frontend/src/store/index.js
schrieveslaach: That's true. I'll fix this.

api/src/infrastructure/docker.rs
schrieveslaach: The code now checks if Docker's REST API return 304 and if so the frontend will received also 202. Other error will be propagated through an 500.

Issue #19, opened 2019-12-04
Deploying companion with custom image removes existing labels and env-vars


dominikks:
Assume I haven an `openid` companion service defined in PREvant's config file as follows:
```
[companions.openid]
serviceName = 'openid'
type = 'application'
image = 'registry.example.com/library/openid:latest'
env = [ "VAR_1=abcd" ]

[companions.openid.labels]
'traefik.frontend.rule' = 'PathPrefix:/{{application.name}}/openid/;'
```

Deploying a custom version of this companion can now be done with the following request to the REST API:
```
POST /api/apps/my-app
[
   {
      "serviceName": "openid",
      "image": "registry.example.com/library/openid:1.0.0"
   }
]
```

This deploys an instance of the `openid` companion with the specified image. However, this instance does not have the labels and environment variables that are defined for this companion in PREvant's config. This means that neither the label `traefik.frontend.rule` nor the environment variable `VAR_1` are set for this instance.

I suggest that PREvant should instead keep the labels and environment variables that are defined in the config file and only override them with values given in the API request.

PR #20, opened 2019-12-05
Merge default and custom configs when deploying a companion

dominikks:
When deploying a custom companion, the configuration for labels, env-vars and volumes is now created by taking the default values (from the source app or the currently running companion) and overwriting the values specified in the deployment request.

Consider the following example. I have a companion defined as follows:
```
[companions.openid]
serviceName = 'openid'
type = 'application'
image = 'private.example.com/library/openid:latest'
env = [ "VAR_1=abcd", "VAR_2=1234" ]

[companions.openid.labels]
'traefik.frontend.rule' = 'PathPrefix:/example.com/openid/;'
'traefik.frontend.priority' = '20000'
```
And I modify it by performing the following request against PREvant's API:
```
POST /api/apps/master
[
  {
    "serviceName": "openid",
    "env": [ "VAR_1=efg" ]
  }
]
```
the service will now run with two env-vars set: `VAR_1=efg` and `VAR_2=1234`. Furthermore, the labels (and volumes, if there are any) are still set.

Labels and Volumes can also be overwritten in a request like env-vars.

Fixes https://github.com/aixigo/PREvant/issues/19.


api/src/apps/apps.rs
```diff
@@ -285,6 +283,46 @@ impl AppsService {
         Ok(services)
     }
 
+    /// Adds all companions to service_configs that are not yet present. Also updates
+    /// service configs if there is a corresponding companion config.
+    fn merge_companions_by_service_name(
+        companions: Vec<ServiceConfig>,
+        service_configs: &mut Vec<ServiceConfig>,
+    ) -> () {
```
schrieveslaach: You can remove `-> ()` here because this is the default in Rust.

api/src/apps/apps.rs
```diff
@@ -285,6 +283,46 @@ impl AppsService {
...
+            service_configs,
+        ));
+    }
+
+    fn merge_service_configs(to: &mut ServiceConfig, from: &ServiceConfig) -> () {
+        // We append the to-envvars to the from-envvars. If the same value
+        // is set twice, docker takes the most recent one.
+        // This way we let docker handle parsing the variables which means
+        // less room for errors.
+        let mut envs = from.env().unwrap_or(&vec![]).clone();
```
schrieveslaach: I would prefer not to rely on Docker here because another infrastructure could reject the deployment. Therefore, the code should prefer the environment variables of the request payload.

api/src/apps/apps.rs
```diff
@@ -285,6 +283,46 @@ impl AppsService {
...
+            }
+        }
+
+        service_configs.extend(AppsService::filter_companions_by_service_name(
+            companions,
+            service_configs,
+        ));
+    }
+
+    fn merge_service_configs(to: &mut ServiceConfig, from: &ServiceConfig) -> () {
```
schrieveslaach: In an other branch I started to put more logic into the `ServiceConfig` itself because I try to avoid to implement anemic domain models. Therefore, I would create a method in `ServiceConfig`. For example:

```rust
impl ServiceConfig {
    pub fn merge_with(&mut self, &other: Self) {
        // ...
    }
}
```

This provides the ability to implement the merge tests there.

api/src/apps/apps.rs
dominikks: The environment variables are currently given as strings of the form `VAR=123` without parsing what the variable and what the allocated value is. As far as I understand this string is given unmodified to Docker without parsing or modifying it. As a consequence, there already is an implicit reliance on Docker with regards to the format of the environment variable string. Some other infrastructure might for example need a format like `VAR '123'` which would be possible with the current implementation, as the string is passe on to the infrastructure without modification.

If we wanted to parse the string however, we would need to define the format supported by PREvant explicitly.

schrieveslaach: That's true. However, I'm already working on it to get rid of this assumption in the Kubernetes branch. In this branch PREvant parses `XXX=YYY` into key and value and it should be merged in the near future. 

Until then you could make here an simple assumption and split by `=` and search for overlapping keys.

dominikks: Alright, I added parsing of environment variables. To avoid breaking compatibility with possible other formats, I added a fallback if an env-string is not of the form `A=B`. I also added tests for the `merge_with` function in `ServiceConfig`. Let me know if you have any more feedback!

dominikks: Alright, I added parsing of environment variables. To avoid breaking compatibility with possible other formats, I added a fallback if an env-string is not of the form `A=B`. I also added tests for the `merge_with` function in `ServiceConfig`. Let me know if you have any more feedback!

api/Cargo.toml
```diff
@@ -18,6 +18,7 @@ handlebars = "1.1"
 hyper = "0.10"
 lazy_static = "1.3"
 log = "0.4"
+maplit = "1.0"
```
schrieveslaach: That should be an development dependency only.

PR #30, opened 2020-02-17
Extend Secret Configuration with Path

### LINKED ISSUE #27: Extend Secret Configuration with Path
PREvant supports the usage of secrets:

```toml
[[services.nginx.secrets]]
name = "key.pem"
data = "LS0tLS1CRUdJTiBFTkNSWVBURUQgUF…JVkFURSBLRVktLS0tLQo="
```

However, these secrets will always live in `/run/secrets/` but container might require a different path. Therefore, the config should support `path` which defaults to `/run/secrets/`. So in following example, `key.pem` will be available at `/opt/key.pem`

```toml
[[services.nginx.secrets]]
name = "key.pem"
data = "LS0tLS1CRUdJTiBFTkNSWVBURUQgUF…JVkFURSBLRVktLS0tLQo="
path = "/opt"
```

schrieveslaach:
Secrets can be configured with an optional path. Fixes #27.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -48,6 +48,9 @@ data = "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS…FUlRJRklDQVRFLS0tLS0K"
 # An optional regular expression that checks if the secret has to be
 # mounted for this service. Default is ".+" (any app)
 appSelector = "master"
+# An optional path that points to the secret's parent directory.
+# Default is "/run/secrets" 
+path = "/run/secrets"
 
 [[services.nginx.secrets]]
 name = "key.pem"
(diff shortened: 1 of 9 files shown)


api/README.md
```diff
@@ -48,6 +48,9 @@ data = "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS…FUlRJRklDQVRFLS0tLS0K"
 # An optional regular expression that checks if the secret has to be
 # mounted for this service. Default is ".+" (any app)
 appSelector = "master"
+# An optional path that points to the secret's parent directory.
+# Default is "/run/secrets" 
+path = "/run/secrets"
```
HelenS52: it would be convenient to also enable a global secrets path. but for now this would work

PR #31, opened 2020-02-20
Fix: Apply Templating With Resolved Port Mapping

schrieveslaach:
PREvant did not include the resolved image ports during the templating because the templating happened before resolving the port mappings. This commit fixes this issue by using a struct that collects the data and it applies the templating as the very last step when it will be converted into the list of service configurations.

### DIFF
--- a/api/src/models/mod.rs
+++ b/api/src/models/mod.rs
@@ -27,7 +27,7 @@
 pub use app_name::{AppName, AppNameError};
 pub use image::Image;
 pub use logs_chunks::LogChunk;
-pub use service::{ServiceBuilder, ServiceBuilderError};
+pub use service::{ContainerType, ServiceBuilder, ServiceBuilderError};
 pub use service_config::{Environment, EnvironmentVariable, Router, ServiceConfig};
 
 mod app_name;
--- a/api/src/services/images_service.rs
+++ b/api/src/services/images_service.rs
@@ -25,12 +25,11 @@
  */
 
 use crate::models::Image;
-use crate::models::ServiceConfig;
 use dkregistry::errors::Error as DKRegistryError;
 use dkregistry::v2::manifest::Manifest;
 use futures::future::join_all;
 use regex::Regex;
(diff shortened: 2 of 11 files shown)


api/src/models/mod.rs
```diff
@@ -27,7 +27,7 @@
 pub use app_name::{AppName, AppNameError};
 pub use image::Image;
 pub use logs_chunks::LogChunk;
-pub use service::{ServiceBuilder, ServiceBuilderError};
+pub use service::{ContainerType, ServiceBuilder, ServiceBuilderError};
```
x1B: Just because I can't seem to find it: Is the additional import used anywhere?

api/src/services/images_service.rs
```diff
@@ -68,27 +67,26 @@ impl ImagesService {
...
         Ok(port_mappings)
     }
 
-    async fn resolve_image_blob<'a>(
-        config: &'a ServiceConfig,
-        image: &Image,
-    ) -> Result<Option<(&'a ServiceConfig, ImageBlob)>, ImagesServiceError> {
-        debug!("Resolve image manifest for {}", config.image().to_string());
+    async fn resolve_image_blob(
+        img: &Image,
```
x1B: Hmm, in other places, `image` is introduced as a name, but here it is renamed to `img`. If this is to avoid Confusion with the `image` below — maybe rename that to `image_name`?

api/src/models/mod.rs
schrieveslaach: It will be used in `crate::apps::DeploymentUnit` and I started to reexport the structs a while ago and `ContainerType` was missing. With this change you can just write `crate::models::ContainerType`.

api/src/services/images_service.rs
schrieveslaach: Done.

schrieveslaach:
@x1B, as we discussed these changes in person I will merge this PR now.

PR #34, opened 2020-02-27
Fix templating of merged companions

dominikks:


api/src/apps/deployment_unit.rs
```diff
@@ -168,30 +131,96 @@ impl TryInto<Vec<ServiceConfig>> for DeploymentUnit {
...
-                &self.app_name,
-                &self.configs[index],
-            )?;
-            configs.push(companion_config);
+        for config in self.configs.into_iter() {
+            services.insert(config.service_name().clone(), config);
+        }
+
+        // Note that if a custom companion is deployed, it will be present in the a
+        // application_companions or service_companions vector as well as in the configs vector.
```
schrieveslaach: I would recommend to write:

> If the user wants to deploy a service that has the same name as a companion, it must be avoided that services will be deployed twice. Therefore, …

api/src/apps/deployment_unit.rs
```diff
@@ -168,30 +131,96 @@ impl TryInto<Vec<ServiceConfig>> for DeploymentUnit {
...
+
+        for companion in service_companions_of_request.iter() {
+            services
+                .get_mut(companion.templated_companion.service_name())
+                .unwrap()
+                .merge_with(&companion.templated_companion);
         }
 
-        configs.extend(self.configs);
+        // Filter out service_companions for deployed service_companions
```
schrieveslaach: Typo: exclude service_companions that are included in the request.

api/src/apps/deployment_unit.rs
```diff
@@ -484,4 +513,92 @@ mod tests {
...
+            vec![sc!("openid", "private.example.com/library/openid:backup")],
+        );
+
+        unit.extend_with_config(&config).unwrap();
+
+        let openid_configs: Vec<_> = unit.try_into().unwrap();
+        assert_eq!(openid_configs.len(), 1);
+        let openid_env = openid_configs[0].env().unwrap();
+
+        assert_eq!(
```
schrieveslaach: you could also assert that `"private.example.com/library/openid:backup"` is the service's image

api/src/apps/deployment_unit.rs
dominikks: How about the following:

> If the user wants to deploy a service that has the same name as a companion, it must be avoided that services will be deployed twice. Furthermore, deploying service companions should be avoided for services that override a service companion.

schrieveslaach: sounds good to me. :+1:

PR #36, opened 2020-03-11
Templated and Replicated Enviroment Variables (Fixes #26)

### LINKED ISSUE #26: Templated Enviroment Variables
Variables are lost when an application is replicated. The variables that a build pipeline explicitly passes may also play an important role for the microservice in the replicated environment. Therefore the build pipeline should be able to define this information through the REST API.

To do this, the `env` field in the service definition should be extended. In #24, the field has already been extended to an object structure. This should be enhanced to support: 

```json
{
   "env": {
        "EXTERNAL_LOGOUT_URL" : {
            "value": "http://....{{application.name}}/",
            "templated": true,
            "replicate": true
       }
   }
}
```

The new parameters `templated` and `replicate` ensure that when replicating a service, PREvant checks whether an environment variable is taken over and if so, whether the handlebars templating mechanism should be applied.

schrieveslaach:
This commit adds support for templated environments. HTTP clients can provide a define that enviroment variables will be templated and they can define that environment variables will be replicated when a service will be replicated.

Closes #35 because this PR use Rust for integration testing.


api-tests/tests/container/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod prevant;
+mod traefik;
+
+pub use prevant::{delete_app, deploy_app, logs, replicate_app, PREvant};
+pub use traefik::{make_reqest, Traefik};
```
HelenS52: make_request

api-tests/tests/docker.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    let mut i = 0;
+    loop {
+        if let Ok(logs) = logs(&prevant, &replicated_app_name, "db").await {
+            if logs.contains("GENERATED ROOT PASSWORD") {
+                break;
+            }
+        };
+
+        std::thread::sleep(Duration::from_secs(15));
```
HelenS52: there is no workaround for thread sleep here?

api/src/config/config.rs
```diff
@@ -106,6 +106,7 @@ impl Config {
         })?)
     }
 
+    // TODO: no result
```
HelenS52: did you forget about this todo?

schrieveslaach: Yes, totally forgot about it. Now it is fixed.

api-tests/tests/docker.rs
schrieveslaach: I wasn't able to get [`futures_retry`](https://docs.rs/futures-retry/0.5.0/futures_retry/) running and there are no good retry-until-APIs I know. Therefore, it was easier to work with sleep.

api/src/models/service_config/enviroment.rs
```diff
@@ -0,0 +1,372 @@
+/*-
```
dominikks: _Typo_ in the filename

api/src/models/service_config/enviroment.rs
```diff
@@ -0,0 +1,372 @@
...
+            .unwrap();
+
+        assert_eq!(e.key, "MYSQL_USER".to_string());
+        assert_eq!(e.value.unsecure(), "admin".to_string());
+    }
+
+    #[test]
+    fn should_parse_env_from_object_without_templating() {
+        let e = from_value::<Environment>(serde_json::json!({
+            "MYSQL_USER": {"value": "admin"}
```
dominikks: Here, I have the same question as above: Do you actually verify that templating was not applied?

api/src/apps/deployment_unit.rs
```diff
@@ -420,6 +416,54 @@ mod tests {
...
+            ))
+        );
+    }
+
+    #[test]
+    fn should_not_apply_templating_on_service_environment_variables() {
+        let mut config = sc!("wordpress-db", "mariadb:10.3.17");
+        config.set_env(Some(Environment::new(vec![EnvironmentVariable::new(
+            String::from("MYSQL_DATABASE"),
+            SecUtf8::from("db"),
```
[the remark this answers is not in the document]
schrieveslaach: True, I will fix this.

api/src/models/service_config/enviroment.rs
schrieveslaach: This test is just about parsing only. It just tests if the JSON is valid and if the deserialized value has the correct content.

api-tests/tests/docker.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        if response.text().await.unwrap().contains("Welcome to nginx!") {
+            break;
+        }
+        std::thread::sleep(Duration::from_secs(5));
+
+        i += 1;
+    }
+
+    assert!(i < 5, "Cannot make request to nginx after {} attempts", i);
```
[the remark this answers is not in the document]
schrieveslaach: Nice catch. I'll move the asserts into the loop.

api/src/models/service_config/enviroment.rs
dominikks: Okay, that makes sense.

PR #39, opened 2020-04-15
Use Guards to Ensure That Only One Request Creates or Deletes an App

schrieveslaach:
This commit ensures that subsequent DELETE requests of the same app ID wait for the results of the first DELETE request. This makes the DELETE idempotent.

### DIFF
--- a/api/src/apps/apps.rs
+++ b/api/src/apps/apps.rs
@@ -36,10 +36,10 @@ use cached::{Cached, SizedCache};
 use chrono::{DateTime, FixedOffset, Utc};
 use handlebars::TemplateRenderError;
 use multimap::MultiMap;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::convert::{From, TryInto};
 use std::str::FromStr;
-use std::sync::Mutex;
+use std::sync::{Arc, Condvar, Mutex};
 use std::time::Duration;
 use tokio::runtime::Runtime;
 use tokio::sync::mpsc;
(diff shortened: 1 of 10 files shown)


api/src/apps/apps.rs
```diff
@@ -894,4 +992,32 @@ Log msg 3 of service-a of app master
...
+
+    #[test]
+    fn should_delete_apps() -> Result<(), AppsServiceError> {
+        let config = Config::default();
+        let infrastructure = Box::new(Dummy::new());
+        let apps = AppsService::new(config, infrastructure)?;
+
+        let app_name = AppName::from_str("master").unwrap();
+        apps.create_or_update(&app_name, None, &service_configs!("service-a"))?;
+        let deleted_services = apps.delete_app(&app_name)?;
```
x1B, line 1004: If I understand correctly, this test does not really cover the complex (racing) case of concurrent requests.
Not sure how difficult it would be to implement that case, as it seems to require multi-threading within the test, so I understand the omission.

api/src/apps/apps.rs
```diff
@@ -48,19 +48,78 @@ use yansi::Paint;
...
-struct DeploymentGuard<'a, 'b> {
-    apps_service: &'a AppsService,
-    app_name: &'b String,
+type GuardedResult = Result<Vec<Service>, AppsServiceError>;
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+enum AppGuardKind {
+    Deployment,
+    Deletion,
+}
```
x1B, line 61: Is it impossible for a deployment to race a deletion? What if the external request A wants to delete an app, and B wants to deploy that app (e.g. triggered by a subsequent build). Maybe it would be simpler to just cover all mutations to the same app using the same guard?

schrieveslaach: That is true that this is an issue but a deployment or deletion should be atomic because they could interfere each other, leading to broken states of the application.

I think that this should be covered in a separate issue, introducing a background thread that deploys/deletes the applications. This thread should use a queue can handle the case of many subsequent deployments/deletions. 

I'll create one and reference it in the code.

api/src/apps/apps.rs
schrieveslaach: There is a second test covering this.

api/src/apps/apps.rs
schrieveslaach: I created #40 for this.

Issue #46, opened 2020-06-10
Separate Caching of Web-Host from HTTP Responder


schrieveslaach:
The caching of the web host meta information of the services is coupled to the HTTP responding thread which causes some problems:

1. The response time is high if PREvant did not resolve any host meta information yet
2. Sometimes the cache object produces a panic:
   ```
   [2020-06-10T05:46:45Z INFO  rocket::rocket] Matched: GET /api/apps application/json (apps)
   thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"', src/apps/apps.rs:162:29
   ```
3. Calling `/tickets` is also slow due to the tight coupling of `Apps::get_apps` and web host resolving. 

Additionally, restarts of PREvant do not recover the previous host meta information (a manual restart of the service is required). 

Therefore, the web host meta resolving should be moved to a background thread that periodically polls the web host meta data. The HTTP responder thread of `/apps` should access the web host meta cache through a Rocket managed state object (`/tickets` won't need that any more).

schrieveslaach:
[`evmap`](https://docs.rs/crate/evmap/10.0.2) should be very suitable for the implementation.

PR #50, opened 2020-08-18
Separate Caching of Web-Host from HTTP Responder

binepedia:
The web host meta resolving is be moved to a background thread (every 5 seconds currently)

Fixes https://github.com/aixigo/PREvant/issues/46

### DIFF
--- a/api/src/apps/apps.rs
+++ b/api/src/apps/apps.rs
@@ -27,29 +27,22 @@
 use crate::apps::DeploymentUnit;
 use crate::config::{Config, ConfigError};
 use crate::infrastructure::Infrastructure;
-use crate::models::request_info::RequestInfo;
 use crate::models::service::{ContainerType, Service, ServiceStatus};
-use crate::models::web_host_meta::WebHostMeta;
-use crate::models::{AppName, LogChunk, ServiceBuilder, ServiceConfig};
+use crate::models::{AppName, LogChunk, ServiceConfig};
 use crate::services::images_service::{ImagesService, ImagesServiceError};
-use cached::{Cached, SizedCache};
-use chrono::{DateTime, FixedOffset, Utc};
+use chrono::{DateTime, FixedOffset};
 use handlebars::TemplateRenderError;
 use multimap::MultiMap;
 use std::collections::{HashMap, HashSet};
 use std::convert::{From, TryInto};
 use std::str::FromStr;
 use std::sync::{Arc, Condvar, Mutex};
-use std::time::Duration;
 use tokio::runtime::Runtime;
-use tokio::sync::mpsc;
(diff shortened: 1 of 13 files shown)


api/src/apps/apps.rs
```diff
@@ -627,9 +452,7 @@ mod tests {
             &service_configs!("service-a"),
         )?;
 
-        let info = RequestInfo::new(Url::parse("http://example.com").unwrap());
-
-        let deployed_apps = apps.get_apps(&info)?;
+        let deployed_apps = apps.get_apps()?;
```
MichaelCkr, line 455: Hier scheint Codeduplizierung vorgelegen zu haben, zumindest musste die gleiche Änderung mehrmals gemacht werden.
Kann man ggf noch ein wenig Duplizierung reduzieren?

api/src/models/request_info.rs
```diff
@@ -35,7 +35,7 @@ pub struct RequestInfo {
 }
 
 impl RequestInfo {
-    #[cfg(test)]
+    // #[cfg(test)]
```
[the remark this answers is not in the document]
schrieveslaach: Seems like remainder of testing. Please, revert.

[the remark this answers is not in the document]
binepedia: Forgot to remove this permanently. This is needed now in meta_host to create the default request_info whose links are later on changed to the real request_infos base url.

api/src/apps/apps.rs
binepedia: These are the unit tests and this is the part for testing, so no duplication in my opinion. here just the method's parameters where changed and so no requestInfo is needed anymore.

api/src/apps/host_meta_cache.rs
```diff
@@ -0,0 +1,189 @@
...
+                Err(error) => {
+                    error!("error: {}", error);
+                    continue;
+                }
+            };
+
+            let resolved_host_meta_infos = runtime.block_on(Self::resolve_host_meta(services));
+
+            info!("resolved infos: {:?}", resolved_host_meta_infos);
+            self.writer.purge();
```
schrieveslaach, line 79: Instead of purging the whole it would be nice to keep existing/valid cache data. This produces a REST request every 5 seconds.

api/src/tickets.rs
```diff
@@ -42,7 +41,6 @@ use std::convert::From;
 pub fn tickets(
     config_state: State<Config>,
     apps_service: State<Apps>,
```
schrieveslaach, line 43: Should be `apps: State<Arc<Apps>>,` otherwise Rocket will answer with 500.

binepedia:
Done in new PR #51

PR #52, opened 2020-08-20
fix of network delay after prevant startup causing meta data problems

binepedia:
fix that network might take a little at startup: sleep at beginning of loop and try for the first minute after startup to fetch meta data even if container is older than 5 min.


MichaelCkr (approved):
Looks not like the best way to check, if an application is running, but this has been similar before this change here.
Probably some event-driven approach would be better in the future.

However, I can see any issues here.

PR #53, opened 2020-11-11
Switch Build of Docker Image to Official Images

schrieveslaach:
The Dockerfile now uses the official Rust images and uses distroless image as a runtime environment. Additionally, this commit updates crate dependencies.


x1B (approved):
LGTM

Issue #40, opened 2020-04-20
Worker Producer Queue for Apps


schrieveslaach:
When multiple builds of a feature branch of a microservice run in parallel most of the builds would fail because PREvant rejects the deployment requests with `409 Conflict`. Instead PREvant could manage a queue of deployment (and deletion) requests and a background task works this off in order by removing a request from the queue and forwarding it to the infrastructure. 

New subsequent request for the same app should be consolidated with requests that haven't been removed from the queue yet. 

This approach enables PREvant to provide an [async API](http://restcookbook.com/Resources/asynchroneous-operations/) so that HTTP clients to not run into HTTP timeouts anymore.

PR #55, opened 2021-02-26
Add async completion API

jpommerening:
~~Just starting with the API description to get the ball rolling and get some early-on support ;)~~

A lot going on here … so … I pulled in 3f83e18 while making the AppsService API async. Then updated [kube](https://github.com/clux/kube-rs) (0.48.0) and [dkregistry](https://github.com/camallo/dkregistry-rs) (some *very* recent git commit) to be compatible with Tokio 1.3.0 and Reqwest 0.11.

Adapting the tests took a while, but all should pass now.

I added rustls for reading PEM certs (kube now wants them already parsed to their binary X.509 representation) and hyper (which was already a transitive dependency) for parsing the `Prefer` header.

----

The `create_app`, `delete_app` and the new `task` endpoint now support `Prefer: respond-async` and `Prefer: respond-async,wait=ts`.

App creation and shutdown now trigger the creation of a temporary container identified by a task id that is present for as long as the operation takes. The new `task` endpoint can be used to "poll" the created container.

Known issues:
- The creation API may respond before the deployment task container is present, so a poll immediately after that might indicate that the task is already completed. The API should probably create the deployment task container synchronously before responding.
- Polling a task and waiting for it's completion produces a different result (`Ok(Vec<Service>)`) than polling a task that is already completed (`NotFound(())`).


jpommerening:
Test script `create-poll.sh`:

```
#!/bin/sh
PAYLOAD='[{"serviceName": "my-service", "image": "docker.io/library/my-image:latest" , "env": {}}]'

WAIT="2"
CONTENT_TYPE="Content-Type: application/json"
PREFER="Prefer: respond-async, wait=$WAIT"
HOST="http://localhost:8000"
URL="$HOST/api/apps/master"

RESPONSE=`curl -v -H "$CONTENT_TYPE" -H "$PREFER" -d "$PAYLOAD" "$URL"`
URL=`echo "$RESPONSE" | jq -re 'if .taskId? and .url? then .url else empty end'`

while test -n "$URL" ; do
  RESPONSE=`curl -v -H "$CONTENT_TYPE" -H "$PREFER" "$URL"`
  URL=`echo "$RESPONSE" | jq -re 'if .taskId? and .url? then .url else empty end'`
done

echo "$RESPONSE"
```

Test script `delete-poll.sh`:

```
#!/bin/sh
WAIT="1"
PREFER="Prefer: respond-async, wait=$WAIT"
HOST="http://localhost:8000"
URL="$HOST/api/apps/master"

RESPONSE=`curl -v -X DELETE -H "$PREFER" "$URL"`
URL=`echo "$RESPONSE" | jq -re 'if .taskId? and .url? then .url else empty end'`

while test -n "$URL" ; do
  RESPONSE=`curl -v -H "$CONTENT_TYPE" -H "$PREFER" "$URL"`
  URL=`echo "$RESPONSE" | jq -re 'if .taskId? and .url? then .url else empty end'`
done

echo "$RESPONSE"
```

api/src/infrastructure/docker.rs
```diff
@@ -499,62 +556,119 @@ impl Infrastructure for DockerInfrastructure {
...
         configs: &Vec<ServiceConfig>,
         container_config: &ContainerConfig,
     ) -> Result<Vec<Service>, Error> {
-        let network_id = self.create_or_get_network_id(app_name).await?;
+        let deployment_container = self
+            .create_deployment_task_container(deployment_id, app_name)
+            .await?;
 
-        self.connect_traefik(&network_id).await?;
+        let result = (async move {
```
schrieveslaach, line 568: You need to help me here: why it is necessary to wrap the code into a `async move` block?

api/src/infrastructure/docker.rs
```diff
@@ -682,6 +796,15 @@ impl Infrastructure for DockerInfrastructure {
     }
 }
 
+/// Helper function to build ContainerFilters
+fn label_filter(label_name: &str, label_value: Option<&String>) -> ContainerFilter {
```
schrieveslaach, line 800: 👍

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -92,7 +92,7 @@ pub enum KubernetesInfrastructureError {
 impl KubernetesInfrastructure {
     pub fn new(
         cluster_endpoint: Url,
-        cluster_ca: Option<Certificate>,
+        cluster_ca: Option<Vec<Vec<u8>>>,
```
schrieveslaach, line 95: I would rather move the parsing of the Der into the configuration object instead of passing bytes

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -303,15 +303,18 @@ impl KubernetesInfrastructure {
         &self,
         app_name: &String,
     ) -> Result<(), KubernetesInfrastructureError> {
-        match Api::v1Namespace(self.client())
+        match Api::<V1Namespace>::all(self.client()?)
             .create(
                 &PostParams::default(),
-                namespace_payload(app_name).into_bytes(),
+                &serde_json::from_value(namespace_payload(app_name))?,
```
schrieveslaach, line 309: IMHO we should move the conversion into `namespace_payload` so that is returns `V1Namespace`. For the moment we could use `expect("with a good comment")` in those payload functions so that we do not have to worry here.

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -661,47 +685,68 @@ impl TryFrom<Deployment> for ServiceBuilder {
...
+        } else {
+            Err(KubernetesInfrastructureError::UnexpectedError {
+                internal_message: String::from("Missing deployment labels, annotations, or both"),
+            })
         }
+    }
+}
 
-        Ok(config)
+impl From<serde_json::Error> for KubernetesInfrastructureError {
```
schrieveslaach, line 745: I think this could be removed if work internally with `k8s_openapi` values instead of strings

api/src/infrastructure/kubernetes/payloads.rs
```diff
@@ -63,15 +63,14 @@ macro_rules! secret_name_from_name {
 }
 
 /// Creates a JSON payload suitable for [Kubernetes' Namespaces](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/)
-pub fn namespace_payload(app_name: &String) -> String {
+pub fn namespace_payload(app_name: &String) -> Value {
```
schrieveslaach, line 66: See comment above: I would prefer to return [`Namespace`](https://docs.rs/k8s-openapi/0.11.0/k8s_openapi/api/core/v1/struct.Namespace.html) instead of `Value`.

api/src/webhooks.rs
```diff
@@ -48,5 +49,5 @@ pub fn webhooks(
     );
 
     let app_name = AppName::from_str(&web_hook_info.get_app_name());
-    delete_app(app_name, apps)
+    delete_app_sync(app_name, apps, tasks)
```
schrieveslaach, line 52: 😎

schrieveslaach (changes requested):
Nice work. 😎 I found some minor issues and we should be able to work them out quickly. 

However, we should test the K8s again but I can help you to test it (write me a private message and we do it together). Unfortunately, we do not have integrations tests for K8s.

api/src/apps/routes.rs
```diff
@@ -259,10 +367,56 @@ impl From<AppsError> for HttpApiProblem {
...
+
 impl<'a> FromFormValue<'a> for AppName {
     type Error = String;
 
     fn from_form_value(form_value: &'a RawStr) -> Result<Self, Self::Error> {
         AppName::from_str(form_value).map_err(|e| format!("{}", e))
     }
 }
+
+impl<'a, 'r> FromRequest<'a, 'r> for RunOptions {
```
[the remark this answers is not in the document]
jpommerening: Ah! Nice catch, that was exactly what I was missing! 👏

api/src/infrastructure/docker.rs
jpommerening: Maybe there's a better way, but the part inside the block handles a bunch of error conditions with the `?` operator.
I wanted to make sure that the temporary container gets deleted even in cases where the function exits early because of an error. I looked at implementing the `Drop` trait or using something like the `scopeguard` crate to take care of that. However, since `delete` is async, both of those don't work (unless find a way to poll the completion synchronously).

api/src/infrastructure/kubernetes/infrastructure.rs
jpommerening: Do you mean changing the signature to `Option<Vec<Der>>`?

I chose to use `Vec<Vec<u8>>` to avoid needing `kube`-specifics (`kube::config::Der`) in the config and because `kube` 0.49 and newer actually expect bytes as input.

I can go ahead though and make the change!

jpommerening:
Thanks for taking the time!
I'll go ahead and make some edits, and will get in contact for testing the k8s stuff 👍

api/res/openapi.yml
```diff
@@ -185,6 +221,36 @@ paths:
             text/plain:
               schema:
                 type: string
+  /tasks:
```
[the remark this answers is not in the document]
jpommerening: That's right, also the description of the `/tasks/{taskId}` responses does not exactly match the implementation yet, but that's partly because of the "known issues" mentioned above (polling completed tasks returns 404).

api/src/infrastructure/kubernetes/infrastructure.rs
jpommerening: This is so much nicer! 🥰
On that note: there's `kube-derive` to generate payload schemas from spec structs: https://docs.rs/kube-derive/0.48.0/kube_derive/derive.CustomResource.html
However, it would kind of a pain to properly specify the schema for Traefik …

For now, I think I'll leave the ingress and middleware payloads as `Value`s.

jpommerening:
Okay, I've got a little update:

I made some adjustments to get the kubernetes infrastructure running … to a degree.
I'm quite certain that the issues I'm still having with it are not caused by the update. I noticed some things:

- Deployments are inspected to get the state of the services, however a k8s deployment spec may not be paused, have replicas >= 1 and still, the pods and containers may not yet be ready. Is this intended? [Source](https://github.com/aixigo/PREvant/blob/983cde422fe07266ac15a5c8342e7266d4f8880c/api/src/infrastructure/kubernetes/infrastructure.rs#L646-L657)
  Result: services are listed as running when, in fact, they are not.
  (`curl localhost:8000/api/apps | jq .$app_name` vs `kubectl --namespace $app_name get deployments`)
- At least on my machine, running Docker for Mac, services can't communicate with each other. The issue might be that each service belonging to an application is in its own deployment and deployments are isolated from each other?
  Result: services won't start up because they are waiting for their database. (An app companion in my case, which I assume is stopping the service instance(s) and some other companions from starting)

api/src/apps/routes.rs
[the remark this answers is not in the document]
jpommerening: Resolved with 8f460ec

api/src/infrastructure/kubernetes/infrastructure.rs
jpommerening: Resolved with db1244 and 9f8d95

jpommerening:
@schrieveslaach I think I addressed all of you notes, except for those regarding the OpenAPI documentation.

Would you mind doing another review/test, while I adjust the docs (and remove the response body for tasks)?

api/res/openapi.yml
```diff
@@ -185,6 +221,36 @@ paths:
...
+      summary: Get the state of the given task
+      description: >-
+
+      responses:
+        '200':
+          description: The state of the task
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TaskCompletionState'
```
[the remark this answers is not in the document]
jpommerening: For the time being, I removed the task response altogether.

I still would like to see it working as initially intended, so that you can get the original response after polling, maybe even after the task completed, but one step at a time, right? :)

So, I described it as it's currently implemented:
- As long as the task endpoint does not return 404, the task is running.
- Once you get a 404, the task is finished.
- The polling resource does not have a response body.

api/res/openapi.yml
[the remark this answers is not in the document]
jpommerening: Done 👍

api/res/openapi.yml
```diff
@@ -113,6 +132,23 @@ paths:
...
+          headers:
+            Location:
+              description: The url of the queued task
+              schema:
+                type: string
+                format: url
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TaskCompletionState'
```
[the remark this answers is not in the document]
jpommerening: Yep, I removed the task response altogether, so there should not be any room for interpretation anymore :+1:

jpommerening:
What's your opinion on https://github.com/aixigo/PREvant/pull/55#discussion_r597489114 and https://github.com/aixigo/PREvant/pull/55#discussion_r597481940 ?

api/src/infrastructure/kubernetes/infrastructure.rs
schrieveslaach: Okay, I see your point.

But, If the code passes just the bytes than the validation if the user has passed a path to valid certificate will happen very late at runtime (ignore the inner workings of `read_ca_file` for now). That might be confusing because he or she has just pointed the configuration to a file that does not contain a valid certificate. Therefore, I would be better if `read_ca_file` ensures that the file contains a valid certificate to inform the user that there is something as soon as possible.

There is now `rustls-pemfile` in use within `read_ca_file` but it is very young and has three commit in it (not happy with that). Therefore, I would recommend to use [`openssl's X509`](https://docs.rs/openssl/0.10.33/openssl/x509/struct.X509.html) because it is a dependency anyway and it gives the X509 type that you can use here. Based on this type you can convert it [`to_der`](https://docs.rs/openssl/0.10.33/openssl/x509/struct.X509.html#method.to_der)

api/src/apps/apps.rs
```diff
@@ -258,34 +275,47 @@ impl AppsService {
...
     }
 
     /// Deletes all services for the given `app_name`.
-    pub fn delete_app(&self, app_name: &AppName) -> Result<Vec<Service>, AppsServiceError> {
+    pub async fn delete_app(
+        &self,
+        app_name: &AppName,
+        deployment_id: &TaskId,
+    ) -> Result<Vec<Service>, AppsServiceError> {
         let guard = self.create_or_get_app_guard(app_name.clone(), AppGuardKind::Deletion)?;
```
schrieveslaach, line 301: I'm wondering if we can remove the guard completely because the infrastructure now handles this with `TaskId`. Therefore, the test `should_delete_apps_from_parallel_threads_returning_the_same_result` could be restructured

api/src/infrastructure/docker.rs
```diff
@@ -444,32 +488,45 @@ impl DockerInfrastructure {
...
+                }
+                Err(err) => {
+                    if first_error.is_none() {
+                        first_error = Some(err);
+                    }
+                }
+            }
         }
 
+        if first_error.is_some() {
```
schrieveslaach, line 527: You can avoid unwrap here:

```rust
if let Some(first_error) = first_error {
   return Err(Error::from(first_error));
}
```

schrieveslaach (changes requested):
Unfortunately, I could not finish the review today (I'll continue tomorrow). So far the PR look promising and I just found some minor issues (see comments).

> The issue seems to be that the ports of the databases can't be resolved, so the k8s services are created on port 80

That's a known issue (at least for me). You have to use a private Docker registry that allows you to resolve the container manifest (hub.docker.io will give you 404 if you request the manifest).  For testing, just push an nginx to a private registry and use the image from there.

api/src/infrastructure/kubernetes/infrastructure.rs
jpommerening: Got it!

fwiw, `rustls-pemfile` is not *completely* without a history, as it was just [spun out of rustls](https://github.com/ctz/rustls#release-history), but OpenSSL is already there, and you just did the hard work for me by looking up the right methods ;)

api/src/apps/routes.rs
```diff
@@ -247,6 +339,7 @@ impl From<AppsError> for HttpApiProblem {
             AppsError::AppIsInDeployment { .. } => StatusCode::CONFLICT,
             AppsError::AppIsInDeletion { .. } => StatusCode::CONFLICT,
             AppsError::InfrastructureError { .. }
+            | AppsError::IOError { .. }
```
schrieveslaach, line 342: That is needed because of an open TODO, right?

api/src/apps/apps.rs
```diff
@@ -348,6 +375,9 @@ pub enum AppsServiceError {
     /// Will be used when the service cannot interact correctly with the infrastructure.
     #[fail(display = "Cannot interact with infrastructure: {}", error)]
     InfrastructureError { error: Arc<failure::Error> },
+    /// TODO: REMOVE ME
```
schrieveslaach, line 378: That should be fixed before merging.

api/src/infrastructure/docker.rs
```diff
@@ -85,6 +85,53 @@ impl DockerInfrastructure {
...
+            })
+            .next())
+    }
+
+    async fn create_deployment_task_container(
+        &self,
+        deployment_id: &String,
+        app_name: &str,
+    ) -> Result<ContainerDetails, Error> {
+        // TODO: what to if there is already an deployment
```
schrieveslaach, line 113: We could resolve this TODO in conjunction with removing the deployment guard. Instead of using the deployment guard in `Apps` we could ask the infrastructure if there is a task running and than reject with 409 as it is the case right now. 

Then it should be safe to return an error here if there is a deployment running.

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -266,52 +274,64 @@ impl KubernetesInfrastructure {
...
-            )?;
+                &middleware_payload(app_name, service_config),
+            )
+            .await?;
 
-        self.client().request_text(request).await?;
+        Ok(())
+    }
 
+    async fn create_crds_if_necessary(
```
schrieveslaach, line 301: 👍

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -612,47 +630,49 @@ impl Infrastructure for KubernetesInfrastructure {
...
-                .map(|replicas| {
-                    if replicas <= 0 {
-                        ServiceStatus::Paused
-                    } else {
-                        ServiceStatus::Running
-                    }
+                .as_ref()
+                .map(|spec| match (spec.paused, spec.replicas) {
+                    (Some(true), _) => ServiceStatus::Paused,
+                    (Some(false), Some(replicas)) if replicas <= 0 => ServiceStatus::Paused,
```
schrieveslaach, line 674: 👍

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -764,24 +798,28 @@ mod tests {
 
             $( annotations.insert(String::from($a_key), $a_value); )*
 
-            Object {
-                types: TypeMeta::default(),
+            V1Deployment {
+                // types: TypeMeta::default(),
```
schrieveslaach, line 802: Nitpick: dead code that you can remove.

api/src/main.rs
```diff
@@ -175,37 +175,31 @@ fn read_token_file(path: &PathBuf) -> Result<SecUtf8, StartUpError> {
...
+        File::open(path).map_err(|err| StartUpError::CannotReadCertificateAuthority {
+            path: String::from(path.to_str().unwrap()),
+            err: format!("{}", err),
+        })?;
 
-    let mut buf = Vec::new();
-    if let Err(e) = f.read_to_end(&mut buf) {
-        return Err(StartUpError::CannotReadCertificateAuthority {
+    let mut buffer = Vec::new();
+    file.read_to_end(&mut buffer)
```
schrieveslaach, line 191: The `?` operator is missing here or am I blind?

schrieveslaach: Please run `cargo clippy` to double check. This will print many warnings because I was not aware of it in the beginning of PREvant (I was a Rust beginner) and the code is not yet idiomatic Rust in all places. But I try to improve with every commit. So maybe you will get some hints for your code changes (you can ignore the rest and please, do not try to apply all hints because otherwise the PR will explode 😅 🙈 ).

api/src/models/task_id.rs
```diff
@@ -0,0 +1,91 @@
...
+ */
+use http_api_problem::{HttpApiProblem, StatusCode};
+use rocket::http::RawStr;
+use rocket::request::FromParam;
+use std::str::FromStr;
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct TaskId(uuid::Uuid);
+
+impl TaskId {
```
schrieveslaach, line 34: Here I'm wondering if we should call it `DeploymentId` because task is soooo generic. And that makes me wondering if the route should be `/deployment/{deploymentId}` instead. What do you think?

schrieveslaach (changes requested):
Overall the PR looks good to me. 👍 There are some minor issues but we will figure them out. 😉

api/src/main.rs
jpommerening: You're right about the `read_to_end()`! I remember the compiler warning me about the `#[must_use]`, but that disappeared when I added the `map_err` even though I forgot the `?` 😰

And … yeah, making Clippy happy would be a nice task for another PR. Seems like a great way to learn :D

api/src/models/task_id.rs
jpommerening: I've got a slight preference for "task". I associate "deployment" either with something longer-lasting, in the sense that k8s uses the term, or with "the act of deploying something", in which case it would be confusing to poll a deployment when undeploying an app.

But there probably *is* a better term than "task", I just can't find one that's less generic but would still match both deploying and undeploying an app?

api/src/infrastructure/docker.rs
jpommerening: I'm a bit afraid to remove the guard in apps, because we don't have the task-thing in kubernetes yet.

It should still be possible to check for a running task here, right?

api/src/main.rs
jpommerening: Okay, I think the `must_use` warning just went by me without me noticing it.

Should we add `#![deny(unused_must_use)]`?

api/src/infrastructure/docker.rs
jpommerening: I added a check that … just errors if there's already a matching container 😅

api/src/apps/apps.rs
jpommerening: ✅

api/src/apps/apps.rs
schrieveslaach: We can postpone this for another PR.

api/res/openapi.yml
```diff
@@ -185,6 +207,29 @@ paths:
             text/plain:
               schema:
                 type: string
+  /tasks/{taskId}:
```
schrieveslaach, line 210: As stated below, to have a better naming `/apps/{appName}/status-changes/{statusId}` would be my choice for naming this resource. In this PR we can ignore `appName` in the route implementation:

```rust
#[get("/<app_name>/status-changes/<status_change_id>", format = "application/json")]
fn task(
    _app_name: Result<AppName, AppNameError>,
    status_change_id: Result<AppStatusChangeId, AppStatusChangeIdError>,
    apps: State<Arc<Apps>>,
    tasks: State<Tasks>,
    options: RunOptions,
) -> Result<AsyncCompletion<Json<Vec<Service>>>, HttpApiProblem> {
   // …
```

I'll create than an issue that we will use the `app_name` and remove the deployment guard.

schrieveslaach (changes requested):
As we discussed in private we want to rename the resource path that has been proposed in this PR. Until then, I'm looking forward to merge this PR.

api/src/infrastructure/docker.rs
schrieveslaach: That should be part of the issue that I'll create (see [here](https://github.com/aixigo/PREvant/pull/55#discussion_r602309415)).

api/src/infrastructure/docker.rs
```diff
@@ -85,6 +85,55 @@ impl DockerInfrastructure {
...
+            deployment_id,
+            app_name
+        );
+        let container_info = containers.create(&options.build()).await;
+        let ci = container_info?;
+
+        let x = containers.get(&ci.id);
+        let y = x.inspect().await;
+
+        Ok(y?)
```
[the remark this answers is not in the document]
jpommerening: Ooops 😅
That was me trying to have breakpoints between those steps while debugging 😭

jpommerening:
Testing on our (p)review host revealed some issues coming from shiplift:
```
ERROR prevant::apps::host_meta_cache] Cannot load apps: Cannot interact with infrastructure: Docker Error: channel closed
```

Further investigation needed 😅

api/res/openapi.yml
```diff
@@ -239,6 +279,19 @@ components:
...
+      description: >-
+        Whether the server should respond without waiting for completion
+        (see [RFC-7240](https://tools.ietf.org/html/rfc7240) for details)
+      required: false
+      schema:
+        enum:
+          - respond-async
+          - respond-async, wait=1
+          - respond-async, wait=5
+          - respond-async, wait=10
```
schrieveslaach, line 294: Shouldn't we move this examples to the example block? `wait` isn't limited to these examples.

api/src/apps/apps.rs
schrieveslaach: We have #40 for this.

api/src/apps/routes.rs
```diff
@@ -54,42 +70,100 @@ fn apps(
...
+    app_name: Result<AppName, AppNameError>,
+    status_id: Result<AppStatusChangeId, AppStatusChangeIdError>,
+    apps: State<Arc<Apps>>,
+    options: RunOptions,
+) -> Result<AsyncCompletion<Json<Vec<Service>>>, HttpApiProblem> {
+    let app_name = app_name?;
+    let status_id = status_id?;
+
+    let apps = apps.clone();
+    let future = async move { apps.wait_for_status_change(&status_id).await };
```
schrieveslaach, line 90: These `async move { … }` are not necessary and you can just write:

```rust
let future = apps.wait_for_status_change(&status_id);
```

In this file are two more blocks that do not need an `async move`.

api/src/infrastructure/docker.rs
```diff
@@ -393,81 +497,77 @@ impl DockerInfrastructure {
...
-                .labels
-                .clone()
-                .unwrap()
-                .get(APP_NAME_LABEL)
-                .unwrap()
-                .clone();
-
-            container_details.insert(app_name, details);
+        for container in container_list.into_iter() {
+            if let Some(details) = not_found_to_none(inspect(container).await)? {
```
schrieveslaach, line 560: 🎆 Hurray

schrieveslaach (changes requested):
I found some small code occurrences to improve. Please, have a look. Additionally, we should keep the number of commits on the main branch small. Code you squash your commit?

api/res/openapi.yml
jpommerening: Yeah, it's kinda hard to model the header in a meaningful way.
Another method I could think of would be:
```
      example:
        respond-async,wait=10
      schema:
        type: array
        uniqueItems: true
        items:
          oneOf:
            - type: string
              description: Respond before the operation is finished
              pattern: ^respond-async$
              example: respond-async
            - type: string
              description: Wait the specified amount of seconds for the operation to finish
              pattern: ^wait=(\d+)$
              example: wait=10
```

Swagger UI can even validate it an provide a semi-useful UI (list of strings):
image

api/src/apps/routes.rs
jpommerening: But then the `&status_id`, owned by the route-fn, might be dropped before `wait_for_status_change` completes. I mean, I could `.clone()` the `status_id`, but is that an improvement? (I'm asking because I don't know :D)

api/res/openapi.yml
schrieveslaach: Looks good to me.

api/src/apps/routes.rs
schrieveslaach: Rust's pinning ensures that for you. The [async-book states the same question](https://rust-lang.github.io/async-book/04_pinning/01_chapter.html):

> However, what happens if we have an async block that uses references?

This reference explains it quite well what is done under `rustc`'s hood

schrieveslaach (approved):

schrieveslaach:
We talked about the remaining comments and we came to the conclusion that the code is good and we tested it in our lab. We will merge it now.

PR #58, opened 2021-04-19
Use the global runtime in tickets.rs

jpommerening:
Does not seem to improve memory usage by itself. But here's the change anyway 🙃


schrieveslaach (approved):
Looks good to me

PR #59, opened 2021-04-19
Upgrade to Tokio 1.5 (might fix memory leak)

jpommerening:
Looks like it has no ill effects.
Memory usage on gcr.io/distroless seems to be unaffected even in combination with #58.


schrieveslaach (approved):
Looks good to me. Thanks.

PR #61, opened 2021-04-20
Sleep between host meta crawls

jpommerening:
This reverts to the previous behavior of sleeping *5 secs between* each iteration of crawling host-meta info.

My PR #55 inadvertently switched to sleeping *at most 5 secs*, or not sleeping at all, in case crawling took longer than 5 seconds, leading to increased CPU and memory consumption.

(See: https://docs.rs/tokio/1.5.0/tokio/time/fn.interval.html#examples)


schrieveslaach:
Nice catch. 👍 😎

schrieveslaach (approved):

PR #70, opened 2021-07-05
Javascript Hooks

### LINKED ISSUE #41: Hooks for Advanced Skripting
This issue is currently just an idea. 

Users of PREvant want to extend configurations of services dynamically. This custom configuration might be based on some team-specific development processes and potentially on some state that needs to be persisted. 

Therefore, PREvant could make use of the scripting engine [rhai](https://github.com/jonathandturner/rhai):

In the configuration PREvant could offer a hooks section:

```toml
[hooks.preDeployment]
script = '/path/to/preDeployment.rhai'
stateful = true

[hooks.postDeletion]
script = '/path/to/postDeletion.rhai'
stateful = true
```

Based on this example PREvant would execute `/path/to/preDeployment.rhai` with app name, service configurations, state, etc. and the script could modify, fitler, the service configurtions and state, before the actual deployment happens.

This could replace the secrets eventually.

schrieveslaach:
This PR introduces the concept of hook to PREvant that can be used to manipulate the deployment before handing it over to the actual infrastructure layer. With hooks in place you are able to add custom logik, such as re-configuring services to use a different database configuration.

Fixes #41


api/README.md
```diff
@@ -104,7 +104,6 @@ The list of available handlebars variables:
 
 - `application`: The companion's application information
   - `name`: The application name
-- `applicationPath`: The root path to the application.
```
schrieveslaach, line 107: This variable never existed. So, I decide to remove it.

api/src/apps/mod.rs
```diff
@@ -2,7 +2,7 @@
  * ========================LICENSE_START=================================
```
schrieveslaach, line 2: Due to a hint of `cargo clippy` the file `api/src/apps/apps.rs` has been renamed to `api/src/apps/mod.rs`. Unfortunately, git does not recognize the diff.

api/README.md
```diff
@@ -144,3 +143,36 @@ The list of available handlebars variables:
...
+
+The hook at `path/to/hook.js` must provide following Javascript function:
+
+```javascript
+function deploymentHook(appName, serviceConfigs) {
+  return serviceConfigs.filter( _ => true );
+}
+```
+
+Note: the templating has been already applied and that the hook has only access to the final value and it cannot modify the original values.
```
binepedia, line 166: something is wrong in this sentence. Did you mean `.. and the hook has ..` or otherwise start it with `Note that the..`

api/src/apps/hooks.rs
```diff
@@ -0,0 +1,567 @@
...
+            Some(mut context) => {
+                Self::register_configs_as_global_property(&mut context, &configs);
+                context.register_global_property(
+                    "appName",
+                    Value::String(app_name.to_string().into()),
+                    Attribute::READONLY,
+                );
+
+                let transformed_configs = context
+                    .eval("deploymentHook(appName, serviceConfigs)")
```
binepedia, line 67: I'm confused why this is calles serviceConfigs, when before all I find is configs. How does this work?

binepedia: got it, its below registered as global I guess. Quite confusing code here 😅

api/README.md
```diff
@@ -144,3 +143,36 @@ The list of available handlebars variables:
   - `name`: The service name which is equivalent to the network alias
   - `port`: The exposed port of the service
   - `type`: The type of service. For example, `instance`, `replica`, `app-companion`, or `service-companion`.
+
+## Hooks
```
binepedia, line 147: I first thought a hook is for a complete deployment but not just for one service per deployment. Maybe this could be made a little clearer here.

binepedia:
The changes seem to be fine. Some changes I do not get because I do not know rust that well but they look like they are not that important, so I believe they work.

I think the hooks could be made a little clearer in the README.

binepedia (approved):

api/src/apps/hooks.rs
schrieveslaach: I agree and that was the tricky part. I created [a PR for Boa](https://github.com/boa-dev/boa/pull/1357/files) that explains how to do it. Maybe, this Boa PR clarifies it for you as well.

api/README.md
schrieveslaach: Yes, it is for a complete deployment. The JS code receives the `appName` and an array of all service configurations. I'll check if I can improve the documentation. Otherwise, do you have a proposal how to clarify it?

api/README.md
schrieveslaach: yes, you are right.

api/README.md
```diff
@@ -144,3 +143,36 @@ The list of available handlebars variables:
...
+
+Hooks can be used to manipulate the deployment before handing it over to actual infrastructure. For example, based on the app name you can decide to reconfigure your services to use a different DBMS so that you are able to verify that your services work with different DBMSs.
```
binepedia, line 149: How about:
```
Hooks can be used to manipulate the deployment before handing it over to actual infrastructure. Hooks are called for each service and can manipulate each services configuration. For example, based on the app name you can decide to reconfigure your services to use a different DBMS so that you are able to verify that your services work with different DBMSs. Or based on the service name you can specify manipulation for a specific service.
```
I do not need a new review for this.

schrieveslaach: > Hooks are called for each service

That's not quite true because a hook is called once with all service configurations as an argument. I'll hope my new version clarifies that a little bit better.

PR #77, opened 2021-12-10
add PartialEq for images because of possible None comparison

binepedia:
Since images could have None values for registry, image_user and image_tag, a PartialEq overload is needed for comparison.


binepedia:
Since images could have None values for registry, image_user and image_tag, a PartialEq overload is needed for comparison.

PR #81, opened 2022-01-14
Deployment Strategy

schrieveslaach:
The companions' configuration provides a new parameter that is called `deploymentStrategy` that allows to define if a companion must beredeployed during a deployment request.

Additionally, replicas will be restarted only if there is an image update available.


api/README.md
```diff
@@ -144,6 +144,23 @@ The list of available handlebars variables:
   - `port`: The exposed port of the service
   - `type`: The type of service. For example, `instance`, `replica`, `app-companion`, or `service-companion`.
 
+### Deployment Strategy
+
+Companions offer different deployment strategies so that a companion could be restart or not under certain conditions. Therefore, PREvant offers following configuration flags:
```
binepedia, line 149: restarted

api/README.md
```diff
@@ -144,6 +144,23 @@ The list of available handlebars variables:
...
+```toml
+[companions.openid]
+type = 'application'
+image = 'private.example.com/library/openid:latest'
+deploymentStrategy = 'redeploy-on-image-update'
+```
+
+`deploymentStrategy` offers following values and if a companion exists for an app following strategy will be applied:
+
+- `redeploy-alway` (_default_): Re-deploys the companion every time there is a new deployment request.
```
binepedia, line 160: redeploy-always

api/README.md
```diff
@@ -144,6 +144,23 @@ The list of available handlebars variables:
...
+- `redeploy-on-image-update`: Re-deploys the companion if there is a new image available.
```
binepedia, line 161: newer image available

api/src/infrastructure/kubernetes/payloads.rs
```diff
@@ -251,6 +248,26 @@ pub fn deployment_payload(
     .expect("Cannot convert value to apps/v1/Deployment")
 }
 
+/// Creates an annotation object so that the underlying pod will be deployed according to its
+/// deployment strategy.
+///
+/// For example, this [popular workaround](https://stackoverflow.com/a/55221174/5088458) will be
```
binepedia, line 254: The information here is very confusing as it only is part of RedeployAlways. I would comment where the workaroud is used, so in line 265 or at least make it clear here, that only that part is meant here. So I would say somethink like:
```
 A [popular workaround](https://stackoverflow.com/a/55221174/5088458)  is applied to ensure that for DeploymentStrategy::RedeployAlways k8s will always update the deployment.
```

binepedia:
I did not find anything that really looks wrong. however I do not understand why on most code parts (service)config is renamed to (deployment)strategy. Maybe I just dont get the changes but I would say the strategy is only one part of the config and so in most parts there should be only additional changes and not that many changes on existing code.
So although I don`t see anything that I would directly reject I just don`t get that huge change and if it at all is necessary for this.

schrieveslaach:
> huge change and if it at all is necessary for this.

The number a changes is due to the change in the `Infrastruture` trait: rather than passing a list of `ServiceConfig`s to the Docker or K8s implementation, PREvant passes a list of `DeploymentStrategy` to the implementations. I did not want to extend `ServiceConfig` with a field `DeploymentStrategy` because `ServiceConfig` is deserializable and I did not want to leak the `DeploymentStrategy` into the REST API.

PR #120, opened 2023-01-23
Deprecated Boa lib replaced with latest boa_engine

samuchila:
Deprecated Boa 0.11 replaced with boa_engine 0.16. 
Changes:

- Value module is replaced with JsValue in boa_engine
- Parse implementation of Context is used to produce the StatementList for iteration.
- Statements.run() is disabled, hence replaced with context.eval()
- FunctionDecl.name() will return a Sym (Symbol data type). Hence interner of the context is used to convert to String.

### DIFF
--- a/api/src/apps/hooks.rs
+++ b/api/src/apps/hooks.rs
@@ -26,10 +26,10 @@
 use crate::apps::{Apps, AppsServiceError};
 use crate::infrastructure::DeploymentStrategy;
 use crate::models::{AppName, ContainerType, Environment, EnvironmentVariable, Image};
-use boa::exec::Executable;
-use boa::property::Attribute;
-use boa::syntax::ast::node::Node;
-use boa::{Context, Value};
+
+use boa_engine::property::Attribute;
+use boa_engine::syntax::ast::node::Node;
+use boa_engine::{Context, JsValue};
 use secstr::SecUtf8;
 use std::collections::BTreeMap;
 use std::iter::IntoIterator;
@@ -58,7 +58,7 @@ impl Apps {
                 Self::register_configs_as_global_property(&mut context, &configs);
                 context.register_global_property(
                     "appName",
-                    Value::String(app_name.to_string().into()),
+                    JsValue::String(app_name.to_string().into()),
                     Attribute::READONLY,
                 );
 
@@ -83,7 +83,8 @@ impl Apps {
             }
         };
 
-        let statements = match boa::parse(&hook_content, false) {
+        let mut context = Context::default();
+        let statements = match context.parse(&hook_content) {
             Ok(statements) => statements,
             Err(err) => {
                 error!("Cannot parse hook file {:?}: {}", hook_path, err);
@@ -91,8 +92,7 @@ impl Apps {
             }
         };
 
-        let mut context = Context::new();
-        if let Err(err) = statements.run(&mut context) {
+        if let Err(err) = context.eval(&hook_content) {
             error!(
                 "Cannot populate hook {:?} to Javascript context: {:?}",
                 hook_path, err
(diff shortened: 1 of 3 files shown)


api/src/apps/hooks.rs
```diff
@@ -83,16 +83,17 @@ impl Apps {
             }
         };
 
-        let statements = match boa::parse(&hook_content, false) {
+        let mut context = Context::default();
+        let statements = match Context::default().parse(&hook_content) {
```
schrieveslaach, line 87: You can avoid an extra allocation here.

```suggestion
        let statements = match context.parse(&hook_content) {
```

api/src/apps/hooks.rs
```diff
@@ -104,7 +105,7 @@ impl Apps {
             .items()
             .iter()
             .find_map(|node| match node {
-                Node::FunctionDecl(decl) if decl.name() == "deploymentHook" => Some(decl),
+                Node::FunctionDecl(decl) if context.interner().resolve_expect(decl.name()) == "deploymentHook" => Some(decl),
```
schrieveslaach, line 108: That might panic.

```suggestion
                Node::FunctionDecl(decl) if context.interner().resolve(decl.name()) == Some("deploymentHook") => Some(decl),
```

api/src/apps/hooks.rs
```diff
@@ -117,7 +118,7 @@ impl Apps {
             .collect::<Vec<_>>();
 
         let js_configs = serde_json::to_value(js_configs).expect("Should be serializable");
-        let js_configs = Value::from_json(js_configs, context);
+        let js_configs = JsValue::from_json(&js_configs, context).unwrap();
```
schrieveslaach, line 121: We should handle the error case and maybe bubble that up.

schrieveslaach (changes requested):
Overall the changes look good to me. Thanks for updating this dependency. I failed in the past.

There are some issues to be fixed. Additionally, you should run `cargo fmt` before commiting any changes.

Thanks in advance.

api/src/apps/hooks.rs
samuchila: Extra allocation removed

api/src/apps/hooks.rs
samuchila: Some() wrapper added

api/src/apps/hooks.rs
samuchila: Error case is handled with below line
let js_configs =
            JsValue::from_json(&js_configs, &mut context).expect("Unable to read JSON value");

api/src/config/mod.rs
```diff
@@ -196,9 +196,7 @@ impl Config {
     }
 
     pub fn hook(&self, hook_name: &str) -> Option<&PathBuf> {
-        self.hooks
-            .as_ref()
-            .and_then(|hooks| hooks.get(hook_name))
+        self.hooks.as_ref().and_then(|hooks| hooks.get(hook_name))
```
schrieveslaach, line 199: That's a change that is not related to PR's contribution. Please, revert.

api/src/apps/hooks.rs
```diff
@@ -252,7 +260,7 @@ mod tests {
         let script = r#"
         function deploymentHook( appName, configs ) {
             if( appName === 'master' ) {
-                return configs.filter( service => service.name !== 'service-b' );
+                return configs.filter( service => service.name !== 'service-c' );
```
schrieveslaach, line 263: Why did you change this test? I would that the tests can be keep as they were before upgrading Boa

schrieveslaach (changes requested):
There are some minor clean up tasks. Please, check them. And squash your commit into one commit. After that I'm going to merge and release the PR.

api/src/config/mod.rs
samuchila: This has been reverted.

api/src/apps/hooks.rs
samuchila: This has been reverted. Commits are squashed.

schrieveslaach:
Thanks a lot for updating this component. @drik98 will be also happy about this update.

PR #126, opened 2023-05-05
Builder Pattern Implementation for DeploymentUnit.

samuchila:
Performed the below changes:
- deployment_unit.rs and hooks.rs are moved to a new folder - deployment
- deployment_unit internal implementation changed to a Builder pattern which ensures the deployment unit follows a sequence of stages and also prevents improper use of deployment_unit during compile time
- hooks.rs now contains its own struct which is called from deployment_unit.rs. Dependency between AppsService and Hooks is now removed instead Hooks now uses config/Config.rs struct.
- Tests for hooks.rs is now refactored as required.


api/src/deployment/deployment_unit.rs
```diff
@@ -0,0 +1,969 @@
...
+
+    #[tokio::test]
+    async fn should_return_unique_images() -> Result<(),AppsServiceError>{
+        let config = Config::default();
+        let unit = DeploymentUnitBuilder::init(AppName::from_str("master").unwrap(), vec![sc!("http1", "nginx:1.13"), sc!("http2", "nginx:1.13")])
+            .extend_with_config(&config)
+            .extend_with_templating_only_service_configs(vec![sc!("http1", "nginx:1.13"), sc!("http2", "nginx:1.13")])
+            .images();
+        
+        assert_eq!(unit.len(), 1);
```
schrieveslaach, line 418: Nitpick: I would test that unit (images would be a better name, BTW) contains "nginx:1.13" only.

api/src/infrastructure/docker.rs
```diff
@@ -24,1325 +24,1328 @@
...
-        assert_eq!(
-            service.config().env().unwrap().get(0).unwrap(),
-            &EnvironmentVariable::with_replicated(
-                String::from("MYSQL_ROOT_PASSWORD"),
-                SecUtf8::from("example")
-            )
-        );
-    }
-}
+ use crate::deployment::deployment_unit::DeploymentUnit;
```
schrieveslaach, line 27: It seems that this file has been wrongly formatted. Please, double check and revert. I would expect that many changes… Only a few lines should have been adjusted.

api/src/infrastructure/dummy_infrastructure.rs
```diff
@@ -24,163 +24,166 @@
...
-    async fn change_status(
-        &self,
-        _app_name: &String,
-        _service_name: &String,
-        _status: ServiceStatus,
-    ) -> Result<Option<Service>, failure::Error> {
-        Ok(None)
-    }
-}
+ use crate::deployment::deployment_unit::DeploymentUnit;
```
schrieveslaach, line 27: Formatting issue

api/src/infrastructure/infrastructure.rs
```diff
@@ -24,118 +24,119 @@
...
-                    .filter(|service| {
-                        *service.container_type() == ContainerType::Instance
-                            || *service.container_type() == ContainerType::Replica
-                    })
-                    .map(|service| service.config().clone())
-                    .collect()
-            }))
-    }
-}
+ use crate::deployment::deployment_unit::DeploymentUnit;
```
schrieveslaach, line 27: Formatting issue

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -31,6 +31,7 @@ use super::payloads::{
     ingress_route_payload, middleware_payload, namespace_payload, secrets_payload, service_payload,
     IngressRoute, Middleware,
 };
+use crate::deployment::deployment_unit::{DeploymentUnit};
```
schrieveslaach, line 34: No braces needed. Please, revisit `cargo fmt`

schrieveslaach (changes requested):
The approach looks very good to me. I made some minor suggestions but you have to fix the formatting first. Something went haywire…

api/src/deployment/deployment_unit.rs
```diff
@@ -0,0 +1,969 @@
...
+    stage: Stage,
+}
+
+pub struct DeploymentUnit {
+    app_name: AppName,
+    _configs: Vec<ServiceConfig>,
+    _service_companions: Vec<(ServiceConfig, crate::config::DeploymentStrategy)>,
+    _app_companions: Vec<(ServiceConfig, crate::config::DeploymentStrategy)>,
+    _templating_only_service_configs: Vec<ServiceConfig>,
+    _image_infos: HashMap<Image, ImageInfo>,
```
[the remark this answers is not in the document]
samuchila: Agreed. The information from the companions and templates are packaged into the strategy. There is no need to expose them. image_info however would be required for future use and not current use. Hence all of them will be removed for now.

api/src/deployment/deployment_unit.rs
samuchila: Yes, the test name states unique. I think we can pass 2 different services and check the uniqueness
vec![
                sc!("http1", "nginx:1.13"),
                sc!("wordpress1", "wordpress:alpine"),
            ]
  If i pass the above config then the test name - should_return_unique_images() will be the right name.

api/src/deployment/deployment_unit.rs
```diff
@@ -0,0 +1,969 @@
...
+            stage: Initialized {
+                app_name: app_name,
+                configs: configs,
+            }
+        }
+    }
+}
+
+impl DeploymentUnitBuilder<Initialized> {
+    pub fn extend_with_config(mut self, config:&Config ) -> DeploymentUnitBuilder<WithCompanions>{
```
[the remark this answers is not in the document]
samuchila: Used rustfmt on the files in commit.

schrieveslaach (approved):
Looks good to me. Thanks for this refactoring.

Issue #123, opened 2023-02-10
Persistent Data


schrieveslaach:
At the moment PREvant does not provide any option to ensure that the data of a service or companion will be persisted. For example, if someone operates a PostgreSQL database on a Kubernetes cluster and the container will be moved to another node, all data will be lost for this application (Kafka is another example).

Further, people report to me personally, that they need a way of sharing data between different services or companions (even when [Microservices shouldn't be synchronized via file system because it is the worst option](https://youtu.be/bbotWkCLuF0?t=355))

Anyways, PREvant should allow that services or companions are able to store data. Here are some ideas:

- The `registry.rs` can be enhanced not just returning the port but also including the list of exposed volumes (according to the [spec](https://github.com/moby/moby/blob/a30990b3c8d0d42280fa501287859e1d2393a951/image/spec/v1.md#image-json-description)
- If configured, PREvant can create volumes in the case of Docker backend and PVCs in the case of Kubernetes backend for companions automatically.
  - What kind of PVC should be configurable for Kubernetes backend
- Hooks should provide a mechanism to share persistent data

samuchila:
@schrieveslaach I have managed to implement persistence for both docker and kubernetes.
I would like to address the second point in the comment above. The best configuration is to create a PVC tied to the `local-path` storage class. (This is created by default when k3s cluster is created, along with local-path-provisioner deployed by default). It has advantages over using host_path and provides an extra degree of flexibility in provisioning.
Additionally, a global config will be used to provide the storage quantity of the provisioned path.

schrieveslaach:
@samuchila, I see you are making progress and I think [my comment here](https://github.com/aixigo/PREvant/pull/127#pullrequestreview-1549215998) should provide some basic structuring to be able to serve all mentioned use cases within this issue which can be implemented first for Docker and then for Kubernetes. Please, have a look and lets focus on Docker first.

PR #127, opened 2023-05-31
Adding Docker Volume Functionality

samuchila:
In a nutshell below is the logic used for implementation
-In infrastructure.rs a list of all available volumes for the app_name is collected
- Once the individual ServiceConfig is accessible, the ImageInfo is extracted from the DeploymentUnit and passed to start_container.
- A combination of app_name, image_info, existing_volume and the service config is passed to a new function to create the host config bind.
- The logic to create host config bind is :
-     If the existing volume does not contain any volume matching the service_name in label then a new volume is
       created in the docker environment,
-     If there is any existing volume then we use `find` functionality to fetch the first occurrence in the list of 
       existing volume. The reason for this being that there should exist only 1 combination of the app_name
       and service_name as labels, No duplicates are allowed.
-     The volume name from any of the above match is merged with attached_volumes(which is the Volumes present 
       in the docker image manifest) in the format "{}:{}"
- The host_config_bind is then passed to the create_container_options function which will ensure the options will result in volume mount creation once container is started.
- A new Container option is added to Enable/Disable feature in config.toml
`[containers]
docker_volume_enabled = false`

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -176,6 +176,22 @@ deploymentStrategy = 'redeploy-on-image-update'
 - `redeploy-on-image-update`: Re-deploys the companion if there is a more rescent image available.
 - `redeploy-never`: Even if there is a new deployment request the companion won't be redeployed and stays running.
 
+### Storage Strategy
+
+Companions may have varying storage requirements and storage strategies cater to these by offering the below configuration flags:
+
+```toml
+[companions.postgres]
+type = 'application'
+image = 'postgres:latest'
+storageStrategy = 'mount-declared-image-volumes'
+```
+
+`storageStrategy` offers following values to determine how storage is managed for a companion:
+
+- `none` (_default_): Companion is deployed without persistent storage.
+- `mount-declared-image-volumes`: Mounts the volume paths declared within the image, providing persistent storage for the companion.
+
 ## Hooks
 
 Hooks can be used to manipulate the deployment before handing it over to actual infrastructure and they are able to manipulate all service configurations once for any deployment REST API call. For example, based on the deployment's app name you can decide to reconfigure your services to use a different DBMS so that you are able to verify that your services work with different DBMSs.
--- a/api/src/infrastructure/docker.rs
+++ b/api/src/infrastructure/docker.rs
@@ -43,15 +43,15 @@ use regex::Regex;
 use shiplift::container::{ContainerCreateInfo, ContainerDetails, ContainerInfo};
 use shiplift::errors::Error as ShipLiftError;
 use shiplift::tty::TtyChunk;
+use shiplift::volume::VolumeInfo;
 use shiplift::{
     ContainerConnectionOptions, ContainerFilter, ContainerListOptions, ContainerOptions, Docker,
-    LogsOptions, NetworkCreateOptions, PullOptions, RegistryAuth,
+    LogsOptions, NetworkCreateOptions, PullOptions, RegistryAuth, VolumeCreateOptions,
 };
 use std::collections::HashMap;
 use std::convert::{From, TryFrom};
 use std::net::{AddrParseError, IpAddr};
 use std::str::FromStr;
-
 static CONTAINER_PORT_LABEL: &str = "traefik.port";
 
 pub struct DockerInfrastructure {
@@ -241,19 +241,37 @@ impl DockerInfrastructure {
         Ok(())
     }
 
+    async fn delete_volume_mount(&self, app_name: &String) -> Result<(), ShipLiftError> {
+        let docker = Docker::new();
+        let docker_volumes = docker.volumes();
+        for volume in DockerInfrastructure::fetch_existing_volumes(app_name).await? {
+            docker_volumes.get(&volume.name).delete().await?
+        }
+        Ok(())
+    }
+
     async fn deploy_services_impl(
         &self,
-        app_name: &String,
-        services: &[DeployableService],
+        deployment_unit: &DeploymentUnit,
         container_config: &ContainerConfig,
     ) -> Result<Vec<Service>, Error> {
+        let app_name = deployment_unit.app_name();
+        let services = deployment_unit.services();
         let network_id = self.create_or_get_network_id(app_name).await?;
 
         self.connect_traefik(&network_id).await?;
-
+        let existing_volumes = DockerInfrastructure::fetch_existing_volumes(app_name).await?;
         let futures = services
             .iter()
-            .map(|service| self.start_container(app_name, &network_id, service, container_config))
+            .map(|service| {
+                self.start_container(
+                    app_name,
+                    &network_id,
+                    service,
+                    container_config,
+                    &existing_volumes,
+                )
+            })
             .collect::<Vec<_>>();
 
         let mut services: Vec<Service> = Vec::new();
@@ -294,6 +312,7 @@ impl DockerInfrastructure {
         }
 
         self.delete_network(app_name).await?;
+        self.delete_volume_mount(app_name).await?;
 
         Ok(services)
     }
@@ -304,6 +323,7 @@ impl DockerInfrastructure {
         network_id: &String,
         service: &DeployableService,
         container_config: &ContainerConfig,
+        existing_volumes: &[VolumeInfo],
     ) -> Result<Service, Error> {
         let docker = Docker::new();
         let containers = docker.containers();
@@ -312,7 +332,6 @@ impl DockerInfrastructure {
         if let Image::Named { .. } = service.image() {
             self.pull_image(app_name, service).await?;
         }
-
         let mut image_to_delete = None;
         if let Some(ref container_info) = self
             .get_app_container(app_name, service.service_name())
@@ -350,7 +369,6 @@ impl DockerInfrastructure {
                     .await?;
             }
             container.delete().await?;
-
             image_to_delete = Some(container_details.image);
         }
 
@@ -362,8 +380,16 @@ impl DockerInfrastructure {
             service.container_type(),
         );
 
-        let options =
-            DockerInfrastructure::create_container_options(app_name, service, container_config);
+        let host_config_binds =
+            DockerInfrastructure::create_host_config_binds(app_name, existing_volumes, service)
+                .await?;
+
+        let options = DockerInfrastructure::create_container_options(
+            app_name,
+            service,
+            container_config,
+            &host_config_binds,
+        );
 
         let container_info = containers.create(&options).await?;
         debug!("Created container: {:?}", container_info);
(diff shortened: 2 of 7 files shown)


api/src/infrastructure/docker.rs
```diff
@@ -399,14 +434,14 @@ impl DockerInfrastructure {
...
         }
-
         Ok(Service::try_from(&container_details)?)
     }
 
     fn create_container_options(
         app_name: &String,
         service_config: &ServiceConfig,
         container_config: &ContainerConfig,
+        host_config_binds: &Vec<String>,
```
schrieveslaach, line 444: `cargo clippy` should recommend to use following syntax (from the top of my head):

```suggestion
        host_config_binds: &[&str],
```

This change make it more usable in more context. 

In the past we that wrong too if you look at the `app_name` parameter. As you are changing the signature, you could fix that for this function too.

api/src/infrastructure/docker.rs
```diff
@@ -450,6 +485,12 @@ impl DockerInfrastructure {
             labels.insert(REPLICATED_ENV_LABEL, replicated_env);
         }
 
+        options.volumes(
```
schrieveslaach, line 488: The suggestion above might save you an iter/collect combination here.

api/src/infrastructure/docker.rs
```diff
@@ -490,6 +531,73 @@ impl DockerInfrastructure {
         Ok(())
     }
 
+    async fn get_existing_volumes(app_name: &String) -> Result<Vec<VolumeInfo>, ShipLiftError> {
```
schrieveslaach, line 534: I'm a bit allergic to function names starting only with the verb get and I hate myself for using it when I'm unable to come up with a better name. Functions starting with get imply that it is something fast because it is a simple getter but here it is call to the Docker backend. What about `fetch_`?

```suggestion
    async fn fetch_existing_volumes(app_name: &String) -> Result<Vec<VolumeInfo>, ShipLiftError> {
```

api/src/infrastructure/docker.rs
```diff
@@ -490,6 +531,73 @@ impl DockerInfrastructure {
...
+                        .as_ref()
+                        .map(|label| label.get(APP_NAME_LABEL) == Some(app_name))
+                        .unwrap_or(false)
+                })
+                .collect::<Vec<VolumeInfo>>()
+        })
+    }
+
+    async fn create_docker_volume(
+        app_name: &String,
```
schrieveslaach, line 550: I suggest to run `cargo clippy` and see what lints are recommended for **new** code

```suggestion
        app_name: &str,
```

api/src/infrastructure/docker.rs
```diff
@@ -490,6 +531,73 @@ impl DockerInfrastructure {
...
+        image_info: Option<&ImageInfo>,
+        existing_volume: &Vec<VolumeInfo>,
+        config: &ServiceConfig,
+    ) -> Result<Vec<String>, ShipLiftError> {
+        let mut host_binds = Vec::new();
+
+        let attached_volume = image_info
+            .and_then(|info| info.attached_volume())
+            .unwrap_or_else(|| Vec::new());
```
schrieveslaach, line 577: Instead of `Vec::new()` you might use [`empty()`](https://doc.rust-lang.org/std/iter/fn.empty.html)

```suggestion
            .and_then(|info| info.attached_volume().iter())
            .unwrap_or_else(|| std::iter::empty());
```

api/src/infrastructure/docker.rs
```diff
@@ -490,6 +531,73 @@ impl DockerInfrastructure {
...
+        volumes
+            .create(&volume_options)
+            .await
+            .map(|volume_info| volume_info.name)
+    }
+
+    async fn create_host_config_binds(
+        app_name: &String,
+        image_info: Option<&ImageInfo>,
+        existing_volume: &Vec<VolumeInfo>,
```
schrieveslaach, line 570: Typo and broader context

```suggestion
        existing_volumes: &[VolumeInfo],
```

api/src/infrastructure/docker.rs
```diff
@@ -1346,4 +1475,121 @@ mod tests {
             )
         );
     }
+
+    #[tokio::test]
+    async fn should_create_volume_mount_for_new_deployment() -> Result<(), ShipLiftError> {
```
schrieveslaach, line 1480: I'm not sure about these tests because it is require to a running Docker backend which may pollute your host system if any fails. For integration testing we have a separate crate in the project. If you want, we can discuss in a video chat.

api/src/registry.rs
```diff
@@ -191,6 +220,18 @@ impl ImageConfig {
             .filter_map(|port| u16::from_str(port.as_str()).ok())
             .min()
     }
+
+    fn attached_volume(&self) -> Option<Vec<String>> {
```
schrieveslaach, line 224: I would go with following

```suggestion
    fn attached_volume(&self) -> Vec<&String> {
```

- There should be no need for cloning…
- An empty vec or a none vec same same space in memory: https://users.rust-lang.org/t/space-time-usage-to-construct-vec-t-vs-option-vec-t/35596
- And I agree with:
  > There is no point in wrapping a collection into an Option just for the sake of it. It only makes code more complex, and introduces additional conditional jumps when you need to access the contents.

api/src/infrastructure/docker.rs
```diff
@@ -303,15 +325,15 @@ impl DockerInfrastructure {
         network_id: &String,
         strategy: &DeploymentStrategy,
         container_config: &ContainerConfig,
+        image_info: Option<&ImageInfo>,
+        existing_volumes: &Vec<VolumeInfo>,
```
schrieveslaach, line 329:
```suggestion
        existing_volumes: &[&VolumeInfo],
```

??

api/src/infrastructure/docker.rs
```diff
@@ -361,8 +384,20 @@ impl DockerInfrastructure {
...
-            DockerInfrastructure::create_container_options(app_name, strategy, container_config);
+        let host_config_binds = DockerInfrastructure::create_host_config_binds(
+            app_name,
+            image_info,
+            existing_volumes,
+            strategy,
+        )
+        .await?;
+
+        let options = DockerInfrastructure::create_container_options(
```
schrieveslaach, line 395: I'm missing the part of deleting the volumes when the app will be shutdown.

schrieveslaach (changes requested):
As a first approach the solution looks okay to me. :+1: However, some part is missing for me: there is no clean up/vacuuming after deleting apps. Additionally, I recommend to `cargo clippy` and check if you are introducing more lints to be fixed.

api/src/infrastructure/docker.rs
samuchila: options.volumes consumes Vec<&str> and host_config_binds are created with format!() which is an owned string. One round of iter+collect is needed to make it to Vec<&str>.

api/src/infrastructure/docker.rs
samuchila: As discussed, delete_volume_mount is part of stop_services_impl and is now triggered with self.delete_volume_mount(app_name)

api/src/registry.rs
samuchila: Clone is removed, Vec<&String> is returned as suggested

schrieveslaach (changes requested):
The code looks pretty good to me. However, I'm wondering if we should merge it because automatically mounting all volumes might lead to some unexpected behavior because files in these volumes disappear. Let's chat about it.

samuchila:
@schrieveslaach Added a new option : `docker_volume_enabled` in global config (config.toml).
The option is then loaded onto ContainerConfig which is available in the infrastructure's deploy functions.
The `fetch_existing_volumes` and  `create_host_config_binds` are completely skipped if the option is false.

api/README.md
```diff
@@ -28,6 +28,9 @@ Create a table `containers` with following options:
 
 # Restrict memory usage of containers
 memory_limit = '1g'
+# Optional field to Enable/Disable automatic docker volume mounting for containers. 
+# Defaults to false if no option provided. (Accepted values - true, false)
+docker_volume_enabled = false
```
schrieveslaach, line 33: First of all, thanks for adding such an option to be backward compatible. But I have some obligations with this option:

- The name suggests it works only in Docker environment but you are working on #123 for all backends. So this name is misleading, IMHO.
- It is a boolean and therefore not very descriptive
- It is a global option and there is no control to enable it for specific services. I would like to see that at companion level.

What do you think of following config sketch?

```toml
[companions.postgres]
type = 'application'
image = 'postgres'

[companions.postgres.volumes]
# Alternative value: MountDeclaredImageVolumes
volumeMounts = 'none'
```

Later `volumeMounts` could be extended so that it accepts a list of paths so that the user has control over which volumes will be mounted. This approach might help in the planning extension of hooks that manipulate volume configurations.

api/src/infrastructure/docker.rs
```diff
@@ -491,6 +538,74 @@ impl DockerInfrastructure {
...
+    async fn create_host_config_binds(
+        app_name: &str,
+        image_info: Option<&ImageInfo>,
+        existing_volume: &[VolumeInfo],
+        config: &ServiceConfig,
+    ) -> Result<Vec<String>, ShipLiftError> {
+        let mut host_binds = Vec::new();
+
+        let attached_volume = match image_info {
+            Some(image) => image.attached_volume(),
```
schrieveslaach, line 583: If we think of the aforementioned configuration approach, we should move the logic which volumes need to created for each service into the `DeploymentUnitBuilder`. This builder should enrich `DeployableService` with a list of volumes to be mounted. 

Later we can extend that in the hooks sections to be modifiable when users of PREvant want to share persistent data across services.

schrieveslaach (changes requested):
I took another look into the PR and I think we should spent some more time to be prepared for more use cases. See comments and let's have another chat.

api/README.md
samuchila: After looking at the code I realized this would cause a conflict. Currently `.volumes` alias is occupied by the `files` variable, meaning straight forward adding under volumes means that `volumeMounts` should also apply to `files`.
I don't think that is what you are looking for.

- [ ] I can add some complexity to handle `volumeMounts` only if it is under `.volumes` and is ignored if under `.files` which according to me gives a better meaning.

- [ ] Alternative approach is to maintain `volumeMounts` as a new variable under companions. Meaning it would be referred to as `companions.postgres.volumeMounts` without any dependency on `volumes` or `files`.

Let me know which approach makes better sense.

samuchila:
@schrieveslaach I have made the minor change that maintains the same behaviour as the master branch when no `host_config_binds` are created. Please review the changes.

schrieveslaach (approved):
Looks very good. Thanks for your contribution.

PR #131, opened 2023-07-02
Goji Updated to Jira_Query

AnumDastgir:
Goji is a Rust interface for Jira which isn't being maintained anymore. So, a more updated and maintained alternative for it is jira_query. So, Goji is replaced with jira_query.
-In order to make it work:
1.  Add username and password for Jira in the config file.


AnumDastgir:
Closing in favor of #132

PR #132, opened 2023-07-11
Substitute current rust interface for Jira=> Goji with jira_query

AnumDastgir:
Goji is a Rust interface for Jira which isn't being maintained anymore. So, a more updated and maintained alternative for it is jira_query. So, Goji is replaced with jira_query.
-In order to make it work:

1. Add username and password for Jira in the config file.

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -24,6 +24,7 @@ figment = { version = "0.10", features = ["env", "toml"] }
 futures = { version = "0.3", features = ["compat"] }
 handlebars = "2"
 http-api-problem = "0.56"
+jira_query = "1.3.0"
 k8s-openapi = { version = "0.18", default-features = false, features = ["v1_24"] }
 kube = { version = "0.82", features = ["derive"] }
 lazy_static = "1.4"
@@ -49,9 +50,6 @@ url = { version = "2.1", features = ["serde"] }
 uuid = { version = "1.1", features = ["serde", "v4"] }
 yansi = "0.5"
 
-[dependencies.goji]
-git = "https://github.com/softprops/goji.git"
-rev = "5916526332951a144b6f6fd7c94d02ea56fdad31"
 
 [dependencies.dkregistry]
 git = "https://github.com/camallo/dkregistry-rs.git"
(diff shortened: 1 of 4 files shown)


api/Cargo.toml
```diff
@@ -43,9 +43,7 @@ url = { version = "2.1", features = ["serde"] }
 uuid = { version = "1.1", features = ["serde", "v4"] }
 yansi = "0.5"
 
-[dependencies.goji]
-git = "https://github.com/softprops/goji.git"
-rev = "5916526332951a144b6f6fd7c94d02ea56fdad31"
+jira_query = "1.2"
```
samuchila, line 46: There is a new version available already. It would be great to use the latest version from the get go.

```suggestion
jira_query = "1.3.0"
```

api/src/models/ticket_info.rs
```diff
@@ -36,19 +35,13 @@ pub struct TicketInfo {
 }
 
 impl From<Issue> for TicketInfo {
-    fn from(issue: Issue) -> Self {
-        let summary = match issue.summary() {
-            None => "".to_owned(),
-            Some(summary) => summary,
-        };
+    fn from(issues: Issue) -> Self {
```
samuchila, line 38: Since we are referring to a single issue, its preferred to call it issue itself.
```suggestion
    fn from(issue: Issue) -> Self {
```

api/src/tickets.rs
samuchila, line 1: This seems like a formatting issue causing the whole file to be shown as changed. Usually only the lines you have touched should be shown in changes.
I would suggest to perform a rustfmt on the file along with clippy for the code you have written only.

api/Cargo.toml
samuchila, line 46: There is a new version available already. It would be great to use the latest version from the get go.
```suggestion
jira_query = "1.3.0"
```

samuchila (changes requested):
Thanks for the contribution @AnumDastgir . The changes looks good. Based on previous interactions with Marc, I have added a few comments. Please try to address them.

api/src/tickets.rs
AnumDastgir: Thanks a lot @samuchila for such constructive suggestions. I'll make the changes suggested.

schrieveslaach (approved):
I looked at the code and I tested your change. Works as before. Thanks a lot.

I approved your PR. However, please squash your commits into one commit while keeping your first commit message. Then I'll merge your PR.

PR #135, opened 2023-08-22
Adding Kubernetes Storage (Persistent Volume) Functionality

samuchila:
Adding new Feature that supports Kubernetes Storage a.k.a Persistent Volume capability to companions. The aim of this approach is to keep the functionality and behaviour as close to Docker Volume implementation as possible using rancher's local-path-provisioner.
- The feature allows end users to assign a persistent volume with the configurations under the parent config `[runtime.storageConfig]`

> storageSize - Storage size reserved for the companions. Default is 2Gi. Multiple formats are supported due to the functionality provided by k8s crate.

> storageClass - Storage Class assigned to the created PVCs. Since PREvant is deployed in a k3s cluster, there is an added advantage of a local-path-provisioner by rancher which is deployed by default. Hence we use this as default in case a manual storage is not given.

- Once the above configs are selected, Kubernetes infrastructure layer of PREvant performs the below operations

> Persistent Volume Claim & Persistent Volume (Only in case of host-path) are created for companions and the details are passed on to the Payload Creation Logic.

> Existing logic that handles volume and volumeMounts for `files` is leveraged to accommodate the persistent volume details.

> Each path declared in the image will be allotted with a PVC and dynamically a Persistent volume is created.

> Cleanup is taken care automatically when namespace is deleted as PV, PVC are tied to the namespace.

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -15,6 +15,7 @@ path = "src/main.rs"
 async-trait = "0.1"
 base64 = "0.21"
 boa_engine = "0.17"
+bytesize = { version = "1.3", features = ["serde"] }
 chrono = { version = "0.4", features = ["serde"] }
 clap = { version = "3.2", features = ["derive", "cargo"] }
 env_logger = "0.10"
--- a/api/README.md
+++ b/api/README.md
@@ -17,6 +17,14 @@ type = 'Kubernetes'
 # This information is crucial if you run PREvant behind a Traefik instance that enforces the user ot be
 # logged in.
 labelsPath = '/run/podinfo/labels'
+
+[runtime.storageConfig]
+# Size of the storage space that is reserved and mounted to the deployed companion with storage.
+# If unspecified storage is defaulted to 2g.
+storageSize = '10g'
+# Storage class denotes the type of storage to be used for companions deployed with storage.
+# Manually managed storage classes can be specified here. If unspecified default storage class will be used.
+storageClass = 'local-path'
 ```
 
 ## Container Options
--- a/api/src/config/runtime.rs
+++ b/api/src/config/runtime.rs
@@ -1,5 +1,3 @@
-use std::path::PathBuf;
-
 /*-
  * ========================LICENSE_START=================================
  * PREvant REST API
@@ -25,6 +23,10 @@ use std::path::PathBuf;
  * THE SOFTWARE.
  * =========================LICENSE_END==================================
  */
+use bytesize::ByteSize;
+use serde::Deserialize;
+use std::path::PathBuf;
+
 #[derive(Clone, Debug, Deserialize, PartialEq)]
 #[serde(tag = "type")]
 pub enum Runtime {
@@ -43,11 +45,17 @@ impl Default for Runtime {
 pub struct KubernetesRuntimeConfig {
     #[serde(default)]
     downward_api: KubernetesDownwardApiConfig,
+    #[serde(default)]
+    storage_config: KubernetesStorageConfig,
 }
 
 impl KubernetesRuntimeConfig {
     pub fn downward_api(&self) -> &KubernetesDownwardApiConfig {
         &self.downward_api
+    }
+
+    pub fn storage_config(&self) -> &KubernetesStorageConfig {
+        &self.storage_config
     }
 }
 
@@ -67,6 +75,37 @@ impl Default for KubernetesDownwardApiConfig {
     fn default() -> Self {
         Self {
             labels_path: PathBuf::from("/run/podinfo/labels"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct KubernetesStorageConfig {
+    #[serde(default = "KubernetesStorageConfig::default_storage_size")]
+    storage_size: ByteSize,
+    storage_class: Option<String>,
+}
+
+impl KubernetesStorageConfig {
+    pub fn storage_size(&self) -> &ByteSize {
+        &self.storage_size
+    }
+
+    pub fn storage_class(&self) -> &Option<String> {
+        &self.storage_class
+    }
+
+    fn default_storage_size() -> ByteSize {
+        ByteSize::gb(2)
+    }
+}
+
+impl Default for KubernetesStorageConfig {
+    fn default() -> Self {
+        Self {
+            storage_size: Self::default_storage_size(),
+            storage_class: None,
         }
     }
 }
@@ -112,7 +151,8 @@ mod tests {
             Runtime::Kubernetes(KubernetesRuntimeConfig {
                 downward_api: KubernetesDownwardApiConfig {
                     labels_path: PathBuf::from("/some/path")
-                }
+                },
+                storage_config: KubernetesStorageConfig::default()
             })
         );
     }
@@ -129,5 +169,28 @@ mod tests {
             config.downward_api.labels_path(),
             &PathBuf::from("/run/podinfo/labels")
         )
+    }
+
+    #[test]
+    fn parse_as_kubernetes_storage_config() {
+        let runtime_toml = r#"
+        type = 'Kubernetes'
+        [storageConfig]
+        storageSize = '10g'
+        storageClass = 'local-path'
+        "#;
+
+        let runtime = toml::de::from_str::<Runtime>(runtime_toml).unwrap();
+
+        assert_eq!(
+            runtime,
+            Runtime::Kubernetes(KubernetesRuntimeConfig {
+                downward_api: KubernetesDownwardApiConfig::default(),
+                storage_config: KubernetesStorageConfig {
+                    storage_size: ByteSize::gb(10),
+                    storage_class: Some(String::from("local-path"))
+                }
+            })
+        );
     }
 }
--- a/api/src/infrastructure/kubernetes/infrastructure.rs
+++ b/api/src/infrastructure/kubernetes/infrastructure.rs
@@ -25,11 +25,12 @@
  */
 use super::super::{
     APP_NAME_LABEL, CONTAINER_TYPE_LABEL, IMAGE_LABEL, REPLICATED_ENV_LABEL, SERVICE_NAME_LABEL,
+    STORAGE_TYPE_LABEL,
 };
 use super::payloads::{
     deployment_payload, deployment_replicas_payload, image_pull_secret_payload,
-    ingress_route_payload, middleware_payload, namespace_payload, secrets_payload, service_payload,
-    IngressRoute, Middleware,
+    ingress_route_payload, middleware_payload, namespace_payload, persistent_volume_claim_payload,
+    secrets_payload, service_payload, IngressRoute, Middleware,
 };
 use crate::config::{Config as PREvantConfig, ContainerConfig, Runtime};
 use crate::deployment::deployment_unit::{DeployableService, DeploymentUnit};
@@ -41,9 +42,11 @@ use async_trait::async_trait;
 use chrono::{DateTime, FixedOffset, Utc};
 use failure::Error;
 use futures::future::join_all;
+use k8s_openapi::api::storage::v1::StorageClass;
 use k8s_openapi::api::{
     apps::v1::Deployment as V1Deployment, core::v1::Namespace as V1Namespace,
-    core::v1::Pod as V1Pod, core::v1::Secret as V1Secret, core::v1::Service as V1Service,
+    core::v1::PersistentVolumeClaim, core::v1::Pod as V1Pod, core::v1::Secret as V1Secret,
+    core::v1::Service as V1Service,
 };
 use kube::{
     api::{Api, DeleteParams, ListParams, LogParams, Patch, PatchParams, PostParams},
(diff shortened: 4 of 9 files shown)


api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -478,6 +495,136 @@ impl KubernetesInfrastructure {
 
         Ok(service)
     }
+
+    async fn create_persistent_volume_claim<'a>(
+        &self,
+        app_name: &str,
+        service: &'a DeployableService,
+    ) -> Result<HashMap<&'a String, PersistentVolumeClaim>, KubernetesInfrastructureError> {
```
schrieveslaach, line 503: I'm thinking that expressing that there is sometime not a value is better.

```suggestion
    ) -> Result<Option<HashMap<&'a String, PersistentVolumeClaim>>, KubernetesInfrastructureError> {
```

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -478,6 +495,136 @@ impl KubernetesInfrastructure {
...
+                };
+
+                match Api::namespaced(client.clone(), app_name)
+                    .create(&PostParams::default(), &persistent_volume_claim)
+                    .await
+                {
+                    Ok(pvc) => {
+                        persistent_volume_map.insert(declared_volume, pvc);
+                    }
+                    Err(e) => {
```
schrieveslaach, line 585: Is there a change to check if the PVC existed before? Because I'm not happy with treating it as an none error if there is something else wring.

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -478,6 +495,136 @@ impl KubernetesInfrastructure {
...
+                        "Found more than 1 Persistent Volume Claim - {:?} for declared image path {} \n Using the first available Persistent Volume Claim - {:?}",
+                        &fetched_pvc,
+                        declared_volume,
+                        &fetched_pvc[0].metadata.name
+                    );
+                }
+                persistent_volume_map.insert(declared_volume, fetched_pvc[0].clone());
+            }
+        }
+        Ok(persistent_volume_map)
```
schrieveslaach, line 602: Please, change the code to avoid a mix of `return Ok(persistent_volume_map)` and this `Ok` at the end…

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -478,6 +495,136 @@ impl KubernetesInfrastructure {
...
+    }
+
+    async fn fetch_default_storage_class(&self) -> Option<String> {
+        let storage_classes: Api<StorageClass> = Api::all(match self.client().await {
+            Ok(client) => client,
```
schrieveslaach, line 607: Is that a typo?

```suggestion
            Ok(storage_classes) => storage_classes,
```

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -478,6 +495,136 @@ impl KubernetesInfrastructure {
...
+                        &fetched_pvc[0].metadata.name
+                    );
+                }
+                persistent_volume_map.insert(declared_volume, fetched_pvc[0].clone());
+            }
+        }
+        Ok(persistent_volume_map)
+    }
+
+    async fn fetch_default_storage_class(&self) -> Option<String> {
```
schrieveslaach, line 605: Why string and not `StorageClass`?

```suggestion
    async fn fetch_default_storage_class(&self) -> Option<StorageClass> {
```

api/src/registry.rs
```diff
@@ -254,7 +254,7 @@ mod tests {
     fn should_return_exposed_port() {
         let blob = serde_json::from_str::<ImageBlob>(
             r#"{
-                "config": {
+                 "config": {
```
schrieveslaach, line 257: No semantic changes, please, revert all changes of this file.

api/src/config/runtime.rs
```diff
@@ -71,6 +79,23 @@ impl Default for KubernetesDownwardApiConfig {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct KubernetesStorageConfig {
+    storage_size: Option<String>,
```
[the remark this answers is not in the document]
samuchila: ByteSize made a simpler implementation. Added the same to container.rs with fallback to old implementation.

api/src/infrastructure/kubernetes/payloads.rs
```diff
@@ -211,9 +215,23 @@ pub fn deployment_payload(
...
+                val.push(persistent_volume_mount_payload(path, pvc));
+            }
+            Some(val)
+        }
+        None => {
+            let mut val = Vec::new();
+            for (path, pvc) in persitent_volume_map {
+                val.push(persistent_volume_mount_payload(path, pvc));
+            }
+            Some(val)
```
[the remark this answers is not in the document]
samuchila: Mistake addressed, If no persistent volume map entries then it remains None as per existing behaviour

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -478,6 +495,136 @@ impl KubernetesInfrastructure {
...
+        let Runtime::Kubernetes(k8s_config) = self.config.runtime_config() else {return Ok(HashMap::new())};
+
+        let storage_size = k8s_config.storage_config().storage_size();
+        let storage_class = match k8s_config.storage_config().storage_class() {
+            Some(sc) => Some(sc.into()),
+            None => self.fetch_default_storage_class().await,
+        };
+
+        if storage_class.is_none() {
+            return Ok(HashMap::new());
```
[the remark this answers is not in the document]
samuchila: Rare occurrence, but possible for non k3d clusters.

api/src/infrastructure/kubernetes/infrastructure.rs
samuchila: Directly implemented Option here instead of Result

api/Cargo.toml
```diff
@@ -15,6 +15,7 @@ path = "src/main.rs"
 async-trait = "0.1"
 base64 = "0.21"
 boa_engine = "0.17"
+bytesize = {version = "1.3.0", features = ["serde"]}
```
schrieveslaach, line 18: All versions here don't specify the patch version. Please, update accordingly.

api/src/config/runtime.rs
```diff
@@ -27,6 +23,11 @@ use serde::Deserialize;
  * THE SOFTWARE.
  * =========================LICENSE_END==================================
  */
+use std::path::PathBuf;
+
```
schrieveslaach, line 27: I think we don't use blanks in the use statements anywhere. Please, double check.

api/src/config/runtime.rs
```diff
@@ -79,21 +80,35 @@ impl Default for KubernetesDownwardApiConfig {
...
+    fn default_storage_size() -> ByteSize {
+        ByteSize::gb(2)
+    }
+}
+
+impl Default for KubernetesStorageConfig {
+    fn default() -> Self {
+        Self {
+            storage_size: ByteSize::gb(2),
```
schrieveslaach, line 108:
```suggestion
            storage_size: Self::default_storage_size(),
```

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -500,20 +494,26 @@ impl KubernetesInfrastructure {
         &self,
         app_name: &str,
         service: &'a DeployableService,
-    ) -> Result<HashMap<&'a String, PersistentVolumeClaim>, KubernetesInfrastructureError> {
-        let client = self.client().await?;
+    ) -> Option<HashMap<&'a String, PersistentVolumeClaim>> {
+        let client = match self.client().await.ok() {
```
schrieveslaach, line 498: Why not using the `?` operator here. Or [`let … else { return … }?`](https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html).

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -500,20 +494,26 @@ impl KubernetesInfrastructure {
...
+            None => self
+                .fetch_default_storage_class()
+                .await
+                .map_or_else(|| None, |sc| sc.metadata.name),
         };
 
-        if storage_class.is_none() {
-            return Ok(HashMap::new());
-        }
+        storage_class.as_ref()?;
```
schrieveslaach, line 516: That is kind of an odd pattern here. I would like to see a change where `storage_class` won't be an Option in the first place rather than having a single line using just the `?` operator.

api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -500,20 +494,26 @@ impl KubernetesInfrastructure {
...
+    ) -> Option<HashMap<&'a String, PersistentVolumeClaim>> {
+        let client = match self.client().await.ok() {
+            Some(val) => val,
+            None => {
+                return None;
+            }
+        };
 
-        let Runtime::Kubernetes(k8s_config) = self.config.runtime_config() else {return Ok(HashMap::new())};
+        let Runtime::Kubernetes(k8s_config) = self.config.runtime_config() else {return None};
```
schrieveslaach, line 505: Isn't that the same and the better version?

```suggestion
        let Runtime::Kubernetes(k8s_config) = self.config.runtime_config()?;
```

api/README.md
```diff
@@ -17,6 +17,13 @@ type = 'Kubernetes'
 # This information is crucial if you run PREvant behind a Traefik instance that enforces the user ot be
 # logged in.
 labelsPath = '/run/podinfo/labels'
+
+[runtime.storageConfig]
+# Size of the storage space that is reserved and mounted to the deployed companion with storage.
+storageSize = '10g'
```
schrieveslaach, line 23: We should document the default value here.

api/src/config/runtime.rs
```diff
@@ -112,6 +152,10 @@ mod tests {
             Runtime::Kubernetes(KubernetesRuntimeConfig {
                 downward_api: KubernetesDownwardApiConfig {
                     labels_path: PathBuf::from("/some/path")
+                },
+                storage_config: KubernetesStorageConfig {
+                    storage_size: ByteSize::gb(2),
+                    storage_class: None,
```
schrieveslaach, line 158:
```suggestion
                storage_config: KubernetesStorageConfig::default(),
```

api/src/config/runtime.rs
```diff
@@ -130,4 +174,29 @@ mod tests {
...
+        "#;
+
+        let runtime = toml::de::from_str::<Runtime>(runtime_toml).unwrap();
+
+        assert_eq!(
+            runtime,
+            Runtime::Kubernetes(KubernetesRuntimeConfig {
+                downward_api: KubernetesDownwardApiConfig {
+                    labels_path: PathBuf::from("/run/podinfo/labels")
+                },
```
schrieveslaach, line 194: If possible.

```suggestion
                downward_api: KubernetesDownwardApiConfig::default(),
```

schrieveslaach (changes requested):
I think the code logic looks good. Just some styling hints. Please, checkout my comments.

api/src/infrastructure/kubernetes/infrastructure.rs
samuchila: `?` is not valid here since `runtime_config()` returns `Runtime`
Also I don't think adding the else part is required here, `KubernetesRuntimeConfig` will either have config mapped value or a default value. I don't think there is a scenario where `KubernetesRuntimeConfig` or `Runtime` does not exist as they are always defaulted.
Please correct me if wrong.

api/Cargo.toml
samuchila: Added the new patched dependencies.

api/src/config/runtime.rs
samuchila: cargo fmt avoids formatting if there is extra line in the use statements. Have manually removed it.

api/src/infrastructure/kubernetes/infrastructure.rs
samuchila: This has been avoided and now storage class is a `&str`

api/README.md
samuchila: Default is 2g and is mentioned in the document now

schrieveslaach (changes requested):
Look good to me. Except the change in Cargo.toml

api/Cargo.toml
```diff
@@ -37,7 +38,7 @@ reqwest = { version = "0.11", features = ["json"] }
 rocket = { version = "0.5.0-rc.3", features = ["json"] }
 schemars = "0.8"
 secstr = { version = "0.5", features = ["serde"] }
-serde = "1.0"
+serde = "1.0.188"
```
schrieveslaach, line 41: It is not just that `bytesize` shouldn't carry any patch number. As [stated in the other reviews](https://github.com/aixigo/PREvant/pull/135#discussion_r1315686976) no dependency should have a patch number. If there is a reason for it, please, document it here.

Check **all** changes within this file.

schrieveslaach (changes requested):
The code changes look fine. But there are two things remaining:

- see comment in Cargo.toml
- The commit message states that it adds Kubernetes support to PREvant. But that isn't true. The commit adds support of “Storage Strategy for the Kubernetes Backend”. You can DM me to chat about how to write that up.

schrieveslaach (approved):

PR #136, opened 2023-09-05
Build and Runtime Stabilizations

schrieveslaach:
This PR ensures that the Docker build uses a coherent Debian version across all Docker build stages (Debian Bookworm will be used as a runtime basis and Rust build basis). Additionally, the Rust backend build uses the native-tls crate instead of the OpenSSL crate which should increase development compatibility across different operating systems.

Last but not least, the parsing of host-meta.json resources ignores malformed datetime values and which ensures that other information will be shown in PREvant's UI (e.g. link to OpenAPI specification).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN npm ci && npm run build
 
 
 # Build Backend
-FROM rust:1 as backend-builder
+FROM rust:1-bookworm as backend-builder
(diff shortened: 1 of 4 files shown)


samuchila (approved):
The checks are successful with no SSL errors. Additionally, on local k3d cluster the same SSL error was reported, but on a fresh build of this branch the error is resolved.

PR #137, opened 2023-09-08
Fixed URL Generation Not Taking Forwarded Header Into Account 

### LINKED ISSUE #109: URL Generation Does Not Take Forwarded Header Into Account
Running PREvant behind HTTPS breaking the URLs in the frontend because the [`RequestInfo` construction](https://github.com/aixigo/PREvant/blob/bc93f9ccce110b79a4de24367106a144a9e37bc0/api/src/models/request_info.rs) does not handle this case.

AnumDastgir:
Added and fixed code which takes into account the forwarded headers specifically x-forwarded-host, x-forwarded-proto and x-forwarded port. The url doesn't break any more and works with the required protocols.

Fixes #109

### DIFF
--- a/api/src/models/request_info.rs
+++ b/api/src/models/request_info.rs
@@ -35,28 +35,6 @@ pub struct RequestInfo {
 }
 
 impl RequestInfo {
-    /// Returns the value for the `host` value of the
-    /// [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) header.
-    pub fn host(&self) -> String {
-        match self.base_url.scheme() {
-            "http" => match self.base_url.port() {
-                None | Some(80) => String::from(self.base_url.host_str().unwrap()),
-                Some(port) => format!("{}:{}", self.base_url.host_str().unwrap(), port),
-            },
-            "https" => match self.base_url.port() {
-                None | Some(443) => String::from(self.base_url.host_str().unwrap()),
-                Some(port) => format!("{}:{}", self.base_url.host_str().unwrap(), port),
-            },
-            _ => String::from(self.base_url.host_str().unwrap()),
-        }
-    }
-
-    /// Returns the value for the `proto` value of the
-    /// [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) header.
-    pub fn scheme(&self) -> &str {
-        self.base_url.scheme()
-    }
-
     pub fn get_base_url(&self) -> &Url {
         &self.base_url
     }
@@ -67,16 +45,32 @@ impl<'r> FromRequest<'r> for RequestInfo {
     type Error = ();
 
     async fn from_request(request: &'r Request<'_>) -> request::Outcome<Self, Self::Error> {
-        let hosts: Vec<_> = request.headers().get("host").collect();
+        let forwarded_host = request.headers().get_one("x-forwarded-host");
+        let forwarded_host = match forwarded_host {
+            Some(host) => host.to_string(),
+            None => match request.headers().get_one("host") {
+                Some(host) => host.to_string(),
+                None => {
+                    return Outcome::Failure((Status::BadRequest, ()));
+                }
+            },
+        };
 
-        if hosts.len() != 1 {
-            return Outcome::Failure((Status::BadRequest, ()));
-        }
+        let forwarded_proto = request
+            .headers()
+            .get_one("x-forwarded-proto")
+            .map(|proto| proto.to_string())
+            .unwrap_or_else(|| String::from("http"));
+        let forwarded_port = request
+            .headers()
+            .get_one("x-forwarded-port")
+            .map(|port| format!(":{}", port.to_string()))
+            .unwrap_or_else(|| String::from(""));
 
-        let url_string = format!("http://{}", hosts[0]);
-        match Url::parse(&url_string) {
-            Ok(url) => return Outcome::Success(RequestInfo { base_url: url }),
-            Err(_) => return Outcome::Failure((Status::BadRequest, ())),
+        let host_url = format!("{}://{}{}", forwarded_proto, forwarded_host, forwarded_port);
+        match Url::parse(&host_url) {
+            Ok(url) => Outcome::Success(RequestInfo { base_url: url }),
+            Err(_) => Outcome::Failure((Status::BadRequest, ())),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


api/src/models/request_info.rs
```diff
@@ -57,6 +58,11 @@ impl RequestInfo {
         self.base_url.scheme()
     }
 
+    // Returns the forwarded port.
+    pub fn port(&self) -> &str {
```
schrieveslaach, line 62: I think `port`, `scheme`, and `host` are methods that aren't used anywhere (I think I see some compiler warnings all the time). Therefore, I would like to remove them.

api/src/models/request_info.rs
```diff
@@ -32,6 +32,7 @@ use url::Url;
 #[derive(Clone)]
 pub struct RequestInfo {
     base_url: Url,
+    forwarded_port: String,
```
schrieveslaach, line 35: I think it isn't necessary to store this value.

api/src/models/request_info.rs
```diff
@@ -67,16 +73,40 @@ impl<'r> FromRequest<'r> for RequestInfo {
     type Error = ();
 
     async fn from_request(request: &'r Request<'_>) -> request::Outcome<Self, Self::Error> {
-        let hosts: Vec<_> = request.headers().get("host").collect();
+        // Checking if "x-forwarded-host" header is present
+        let forwarded_host = request.headers().get_one("x-forwarded-host");
+        let host = match forwarded_host {
+            Some(proto) => proto.to_string(),
```
schrieveslaach, line 79: Typo?

```suggestion
            Some(host) => host.to_string(),
```

api/src/models/request_info.rs
```diff
@@ -67,16 +73,40 @@ impl<'r> FromRequest<'r> for RequestInfo {
     type Error = ();
 
     async fn from_request(request: &'r Request<'_>) -> request::Outcome<Self, Self::Error> {
-        let hosts: Vec<_> = request.headers().get("host").collect();
+        // Checking if "x-forwarded-host" header is present
```
schrieveslaach, line 76: I think the code tells the reader exactly what it does.

```suggestion
```

Most of the comments do. Please, adjust accordingly and only keep comments that document something that isn't obvious from code.

api/src/models/request_info.rs
```diff
@@ -67,16 +73,40 @@ impl<'r> FromRequest<'r> for RequestInfo {
...
+        let forwarded_host = request.headers().get_one("x-forwarded-host");
+        let host = match forwarded_host {
+            Some(proto) => proto.to_string(),
+            None => request.headers().get_one("host").unwrap().to_string(), // Default to "host" if not present
+        };
 
-        if hosts.len() != 1 {
+        if host == "" {
```
schrieveslaach, line 83: Due to `unwarp().to_string()` in line 80 the code my fail with a panic in the server and the server wouldn't respond with bad request. I would remove this if block and adjust the match above like this:

```rust
        let forwarded_host = request.headers().get_one("x-forwarded-host");
        let host = match forwarded_host {
            Some(host) => host.to_string(),
            None => match request.headers().get_one("host") {
                Some(host) => host.to_string(),
                None => {
                    return Outcome::Failure((Status::BadRequest, ()));
                }
            }
        }
```

This has the advantage of saving a little CPU cycles (because there is no string allocation if there is no host header) and it doesn't panic.

api/src/models/request_info.rs
```diff
@@ -67,16 +73,40 @@ impl<'r> FromRequest<'r> for RequestInfo {
...
-        let url_string = format!("http://{}", hosts[0]);
-        match Url::parse(&url_string) {
-            Ok(url) => return Outcome::Success(RequestInfo { base_url: url }),
-            Err(_) => return Outcome::Failure((Status::BadRequest, ())),
+        // Checking if "x-forwarded-proto" header is present
+        let forwarded_proto_header = request.headers().get_one("x-forwarded-proto");
+        let forwarded_proto = match forwarded_proto_header {
+            Some(proto) => proto.to_string(),
+            None => "http".to_string(), // Default to "http" if not present
+        };
```
schrieveslaach, line 92: I would rewrite this and the match below and use would use [`unwrap_or_else`](https://api.rocket.rs/v0.5-rc/rocket/mtls/oid/asn1_rs/nom/lib/std/option/enum.Option.html#method.unwrap_or_else) because that makes these `// default to … if not present` comments obsolete.

```suggestion
        let forwarded_proto = request.headers().get_one("x-forwarded-proto").map(|proto| proto.to_string()).unwrap_or_else(|| String::from("http"));
```

schrieveslaach (changes requested):
Thanks for tackling #109 and sorry for the confusion with the forwarded vs x-forwarded-* headers.

The approach looks good to me. I suggested minor code improvements. Please, check out my comments.

I was wondering if we should add some [test via Rocket](https://rocket.rs/v0.5-rc/guide/testing/#testing) to ensure that everything keeps working. Ping me via DM to discuss that.

examples/Kubernetes/PREvant.yml
```diff
@@ -65,6 +65,7 @@ metadata:
 spec:
   entryPoints:
     - http
+    - websecure
```
schrieveslaach, line 68: Please, remove that from the examples because it might not work for others.

```suggestion
```

api/src/models/request_info.rs
```diff
@@ -67,16 +45,33 @@ impl<'r> FromRequest<'r> for RequestInfo {
...
+            .headers()
+            .get_one("x-forwarded-port")
+            .map(|port| format!(":{}", port.to_string()))
+            .unwrap_or_else(|| String::from(""));
 
-        let url_string = format!("http://{}", hosts[0]);
-        match Url::parse(&url_string) {
-            Ok(url) => return Outcome::Success(RequestInfo { base_url: url }),
-            Err(_) => return Outcome::Failure((Status::BadRequest, ())),
+        // Constructing URL based on  headers
```
schrieveslaach, line 70:
```suggestion
```

schrieveslaach (changes requested):
Changes look fine and I appreciate fixing this bug. Please, clean up the suggestion.

AnumDastgir:
> Changes look fine and I appreciate fixing this bug. Please, clean up the suggestion.

Clean up is done.

schrieveslaach (approved):
Looks good to me now. Please, squash your commit into one before merging.

PR #143, opened 2023-11-27
Persistence Hooks for Storage Path Management

### LINKED ISSUE #123: Persistent Data

samuchila:
This feature will allow users to share declared volumes and delete volume paths exposed by the docker manifest with the help of Persistence Hooks.

The hooks also allows adding of any additional volume paths that are undeclared in the image.

Interface
- Input is array service config containing of name and path.
- Functions like sharing persistence paths, adding missing paths, removing existing paths can be performed in the JavaScript Hooks.
- Return from JavaScript to PREvant will be an array of an array where array can contain multiple elements in case of shared and a single element array in case of an individual persistence paths.
- These results are then handled by infrastructure according to the grouping.

Fixes #123

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -200,11 +200,11 @@ storageStrategy = 'mount-declared-image-volumes'
 - `none` (_default_): Companion is deployed without persistent storage.
 - `mount-declared-image-volumes`: Mounts the volume paths declared within the image, providing persistent storage for the companion.
 
-## Hooks
+## Deployment Hooks
 
-Hooks can be used to manipulate the deployment before handing it over to actual infrastructure and they are able to manipulate all service configurations once for any deployment REST API call. For example, based on the deployment's app name you can decide to reconfigure your services to use a different DBMS so that you are able to verify that your services work with different DBMSs.
+Deployment Hooks can be used to manipulate the deployment before handing it over to actual infrastructure and they are able to manipulate all service configurations once for any deployment REST API call. For example, based on the deployment's app name you can decide to reconfigure your services to use a different DBMS so that you are able to verify that your services work with different DBMSs.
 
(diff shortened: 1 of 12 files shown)


samuchila:
@schrieveslaach Request you to please perform the review.

NOTE: One of the previous PR #135 is not merged, causing this change to be shown as a huge one.

schrieveslaach:
> @schrieveslaach Request you to please perform the review.
> 
> NOTE: One of the previous PR #135 is not merged, causing this change to be shown as a huge one.

@samuchila, do you mind to rebase this PR so that it is showing only the relevant changes?

schrieveslaach:
Just thinking out loud: I was wondering if this might be obsolete if we got #146 implemented. Let's chat

schrieveslaach:
As we discussed outside of GitHub, the bootstrapping of environments supersedes this approach. Thus I'm closing this PR in favor of #156.

PR #154, opened 2024-01-09
Add Support for Multi-Platform Image Manifests

### LINKED ISSUE #125: Support Multi-Arch Images
When images are available in [multi-arch](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) PREvant fails to resolve the images' manifest information. Then, the deployed container won't be reachable if it is running on another port than 80 because the exposed port should have been resolved from the manifest.

So, PREvant should be able to handle images with multi-arch builds. Additionally, if the client sends an Image that's manifest couldn't be resolved it should be rejected with an HTTP client error.

schrieveslaach:
This commit add support for retrieving image manifest if the images provides multiple platform variants, e.g. amd64, arm64, …. It also provide better HTTP responses if the API client provided an image that doesn't exist, if the authentication for the image doesn't work or if anything else failed for the image manifest retrieval.

Fixes #125

### DIFF
--- a/api/res/openapi.yml
+++ b/api/res/openapi.yml
@@ -72,7 +72,13 @@ paths:
           description: The application name that will be used to replicate from.
         - $ref: '#/components/parameters/preferAsync'
       requestBody:
-        description: Information of review app to create
+        description: >
+           A list of service configuration that will be included in the application.
+           <p>
+           The example provides two services: a database and a wordpress
+           instance which can also be replicated via the UI. Make sure that
+           you adjust the values for `WP_HOME` and `WP_SITE` to match the
+           host URL of your PREvant instance.
         required: true
         content:
           application/json:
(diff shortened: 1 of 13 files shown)


api/src/registry.rs
```diff
@@ -50,93 +51,100 @@ impl<'a> ImagesService<'a> {
...
+                    // each platform. If a port mapping or a volume definition is different for
+                    // different platforms, that would cripple into issues not just for PREvant but
+                    // rather for all users that migrate to a different architecture.
+                    entries.first().map(|e| e.digest.clone()),
+                )
+            })),
+            ..Default::default()
+        });
+
+        let reference = Reference::from_str(&image.to_string()).unwrap();
```
samuchila, line 121: I think it would be safe to handle the error that Reference::from_str produces. 
I passed image as postgres_ and private-registry.example.com/_/postgres and got the following error:
thread 'rocket-worker-thread' panicked at src/registry.rs:121:65:
called `Result::unwrap()` on an `Err` value: ReferenceInvalidFormat

Problem is that once this has panicked, the tile goes in conflict and gets hard to recover.

api/src/registry.rs
```diff
@@ -215,35 +223,22 @@ impl ImageConfig {
...
-    }
-}
-
-impl From<serde_json::Error> for ImagesServiceError {
-    fn from(e: serde_json::Error) -> Self {
-        ImagesServiceError::InvalidImageBlob {
-            internal_message: format!("{}", e),
-        }
-    }
+pub enum RegistryError {
```
samuchila, line 226: Also on a general note when you launch the application on cargo run -- --runtime-type Kubernetes and start a new app with any payload like below.
`[
  {
    "serviceName": "postgres",
    "image": "postgres:latest",
    "env": {
      "MYSQL_USER": "admin"
    },
    "files": {
      "/etc/mysql/my.cnf": "[client-server]\n# Uncomment these if you want to use a nonstandard connection to MariaDB\n#socket=/tmp/mysql.sock\n#port=3306\n"
    }
  }
]`

Kube returns a `[2024-01-10T11:06:31Z WARN  kube_client::client] Unsuccessful data error parse: 404 page not found`
Which is translated by Routes as an Error but the app is started anyways.
`[2024-01-10T11:06:31Z ERROR prevant::apps::routes] Internal server error: Cannot interact with infrastructure: Unexpected Kubernetes interaction error: ApiError: "404 page not found\n": Failed to parse error data (ErrorResponse { status: "404 Not Found", message: "\"404 page not found\\n\"", reason: "Failed to parse error data", code: 404 })`

This could be slightly misleading.

samuchila (commented):
I have found a couple of minor issues when testing out the changes. @schrieveslaach please assess if these are valid scenarios and address them accordingly.

api/src/registry.rs
schrieveslaach: Nice catch. I included an additional check in `image.rs` which make sure you never reach that point.

api/src/registry.rs
schrieveslaach: I couldn't reproduce this issue.

schrieveslaach: However, I included a working example in the payload. Maybe, you want to check that out as well.

schrieveslaach: Fixed.

api/res/openapi.yml
```diff
@@ -72,14 +72,69 @@ paths:
           description: The application name that will be used to replicate from.
         - $ref: '#/components/parameters/preferAsync'
       requestBody:
-        description: Information of review app to create
+        description: >
+           A list of service configuration that will be included in the application.
+           <p>
+           The example provides two services: a database and a wordpress
+           instance which can also be replicated via the UI. Make sure thta
```
samuchila, line 79: Small typo here -> `thta`

samuchila (approved):
Please consider this as approval as the changes are working without any issues.
In the final commit, please address the minor typo comment.

schrieveslaach:
@samuchila, thanks for your feedback.

Issue #23, opened 2020-01-09
Migrate to Traefik 2


schrieveslaach:
Currently, PREvant supports Traefik 1.7 but Traefik 2.x has been released. Unfortunately, 2.x introduces a new way to define routes, etc. 

In this sense PREvant should replace the companion's labels with the corresponding concepts of [routes](https://docs.traefik.io/routing/overview/), [middlewares](https://docs.traefik.io/middlewares/overview/), etc.

HelenS52:
it would be nice if your fix could also solve the current issue that the networks don't reconnect with traefik

schrieveslaach:
When migration to Traefik 2.x for the Docker Backend take the results of #108 into account.

schrieveslaach:
With version 3 of Traefik the issue became also more apparent  in Kubernetes (see [this line](https://github.com/aixigo/PREvant/pull/226/files#diff-4761c5069b46ce96f49d74476da61f43ced3abc73d22254cd95a2be0e250a593R81) that came via #226):

> currently PREvant supports only Traefik v2, thus the k3s version is pinned to a version that brings Traefik v2.

PR #158, opened 2024-01-18
Use maintained rust-toolchain instead of actions-rs

jannic:
actions-rs seems to be abandoned (https://github.com/actions-rs) and uses deprecated features which cause warnings in github actions. (Like "The following actions uses node12 which is deprecated and will be forced to run on node16: actions-rs/toolchain@v1")

Use dtolnay/rust-toolchain instead of actions-rs/toolchain, and replace actions-rs/cargo by just calling cargo directly.

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Install latest stable
-        uses: dtolnay/rust-toolchain@stable
+        uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Run cargo test
         run:
@@ -26,7 +26,7 @@ jobs:
         run: docker build --pull -t aixigo/prevant .
 
       - name: Install latest stable
-        uses: dtolnay/rust-toolchain@stable
+        uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Run cargo test
         run:
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 
 name: Build and test API
 
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -11,6 +11,8 @@ jobs:
 
       - name: Install latest stable
         uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          rustflags: ""
 
       - name: Run cargo test
         run:
@@ -27,6 +29,8 @@ jobs:
 
       - name: Install latest stable
         uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          rustflags: ""
 
       - name: Run cargo test
         run:
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -10,16 +10,11 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Install latest stable
-        uses: actions-rs/toolchain@v1
-        with:
-            toolchain: stable
-            override: true
+        uses: dtolnay/rust-toolchain@stable
 
       - name: Run cargo test
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --manifest-path api/Cargo.toml
+        run:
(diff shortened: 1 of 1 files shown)


.github/workflows/api.yml
```diff
@@ -10,16 +10,11 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Install latest stable
-        uses: actions-rs/toolchain@v1
-        with:
-            toolchain: stable
-            override: true
+        uses: dtolnay/rust-toolchain@stable
```
schrieveslaach, line 13: I'm not sure what is the best action at the moment. I looked at [nushell's workflow](https://github.com/nushell/nushell/blob/5d63f47c851ff23cdd78e9782372b23d1ab5ae89/.github/workflows/ci.yml#L44C15-L44C52) and they use this one [here](https://github.com/actions-rust-lang/setup-rust-toolchain) (which is also written in Rust).

Do you have any experience with both or one of the actions? Due to the [usage number of actions-rust-lang/setup-rust-toolchain](https://github.com/actions-rust-lang/setup-rust-toolchain/network/dependents) I'm in favor of using actions-rust-lang/setup-rust-toolchain.

schrieveslaach (changes requested):
Thanks for providing this PR and raising the awareness, that action-rs is deprecated now. I'm in favor of switching to a maintained toolchain but I'm uncertain which one (see comment). Maybe you can comment or change the toolchain.

Additionally, do you mind to adjust the `on` attribute to `on: [push, pull_request]` so that we can run tests in PRs as well?

.github/workflows/api.yml
jannic: I have introduced `dtolnay/rust-toolchain` in several repositories related to rust-embedded, like
https://github.com/rust-embedded/embedded-hal/commit/ed9e0c13f134abed81225ec24f15d9cca996378f, about a year ago. I'm not aware of any issues, it "just works".

One interesting design decision with `dtolnay/rust-toolchain` is that one doesn't use a fixed version. Instead, the version specifier `@stable` references a tag (or branch?) that gets updated when new versions of `dtolnay/rust-toolchain` are released. That's of course questionable in terms of reproducibility. On the other hand, one can argue that this whole action is about interacting with external tooling that's not part of the current repository anyway, so having it automatically upgraded when necessary can make the build pipeline more stable.
And indeed, have a look at the history of nushell's workflow you linked to: `actions-rust-lang/setup-rust-toolchain` gets a new version every few months.

Without having used `actions-rust-lang/setup-rust-toolchain` I can't really compare both. I have no reason to believe that one or the other is better.

schrieveslaach: I had another look into `actions-rust-lang/setup-rust-toolchain` and it says on the README:

> The action is heavily inspired by dtolnay's https://github.com/dtolnay/rust-toolchain and extends it with further features.

They also show the usage and there seems to be a moving target things:

> ```yaml
> - uses: actions-rust-lang/setup-rust-toolchain@v1
> ```

So it seems to me that using `actions-rust-lang/setup-rust-toolchain@v1` should do the trick.

Let's do it with `actions-rust-lang/setup-rust-toolchain@v1` and `on: [push, pull_request]`

BTW: with PREvant I try to follow the same strategy: 

> One interesting design decision with dtolnay/rust-toolchain is that one doesn't use a fixed version.

PREvant should stay backwards compatible all the time. Probably, that should be stated in the README. 🤔

jannic: As mentioned in the commit message of https://github.com/aixigo/PREvant/pull/158/commits/7d3f9d543340307b13b85657e4aaed5fd2b34750, simply using actions-rust-lang/setup-rust-toolchain@v1 fails because it automatically sets `RUSTFLAGS = "-D warnings"` to deny all warnings.

@schrieveslaach do you want to keep that and fix all warnings before merging this, or should I manually set `rustflags` to an empty string?

https://github.com/actions-rust-lang/setup-rust-toolchain#rustflags

schrieveslaach: There are some warnings in there due to deprecation attributes duet #23. So empty `RUSTFLAGS` would be best at the moment.

schrieveslaach: Thanks a lot.

schrieveslaach (approved):

PR #156, opened 2024-01-11
Bootstrap Companions in Kubernetes Environments

schrieveslaach:
This commit introduces the configuration `bootstrapping.containers` which provides a way to parse the configuration of application wide companions because the current available configuration of companions is quite limiting (Current backends, Docker and Kubernetes, offer way more options than the PREvant configuration object allows). For example, PREvant was limited to self-contained applications where each microservice only relies on interactions via network API calls (REST, database connections, messaging, etc.) With this commit PREvant is now able to deploy application companions that are more powerful than the PREvant configuration in Kubernetes backends.

If `bootstrapping.containers` is defined, PREvant will start one or more containers on the infrastructure backend that are expected to generate Kubernetes manifests as output on standard out (stdout) that will be parsed by PREvant and supported are:

 - roles and role bindings
 - config maps and secrets
 - service accounts
 - persistent volume claims
 - services
 - pods, deployments, stateful sets, and jobs

Then before deploying these manifests PREvant merges all objects with the objects generated from the HTTP request payload. Thus you can add or overwrite configurations. For example, you can change the image used or an environment variable. If you overwrite any configuration the companion will be turned into an instance (as PREvant did before).

Ingresses won't be deployed if the bootstrap container outputs one of these. Instead they will be parsed and if they use the ingress class `nginx` they will be transformed into Traefik ingresses and middlewares so that the microservices will be available via web interface.

This approach make #143 obsolete and fixes #123 and contributes to #146.

PR #159, opened 2024-01-23
Avoid Duplicated CI Runs on PRs

schrieveslaach:
In #158 the CI workflow has been improved so that CI also runs when providing a PR. However, that led to the situation that CI runs are duplicated for PRs that originate from internal branches. This commit adresses this issues and avoids duplicate CI runs.

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -1,4 +1,8 @@
-on: [push, pull_request]
+on:
+   push:
+      branches:
+         - master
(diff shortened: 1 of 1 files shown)


schrieveslaach:
@jannic, do you mind to have a look. In #156 you can see the effect:

image

jannic (approved):

schrieveslaach:
@jannic, thanks for your feedback.

PR #162, opened 2024-02-27
Improve Log Request Content-Negotiation

schrieveslaach:
The Content-Negotiation for getting logs of services has been improved so that browsers that don't send `text/plain` in the accept header are able to download the logs from the PREvant UI.

### DIFF
--- a/api/src/apps/routes/logs.rs
+++ b/api/src/apps/routes/logs.rs
@@ -1,0 +1,374 @@
+use crate::{
+    apps::Apps,
+    http_result::HttpResult,
+    models::{AppName, AppNameError, LogChunk},
+};
+use chrono::DateTime;
+use futures::stream::StreamExt;
+use http_api_problem::HttpApiProblem;
+use reqwest::header::{ACCEPT, CONTENT_DISPOSITION, LINK};
+use rocket::{
+    http::{Accept, ContentType, RawStr, Status},
+    request::FromRequest,
+    response::stream::{Event, EventStream},
+    response::{Responder, Response},
+    Request, State,
+};
+use std::{str::FromStr, sync::Arc};
+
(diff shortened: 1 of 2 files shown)


api/src/apps/routes/logs.rs
```diff
@@ -0,0 +1,374 @@
...
+    format = "text/event-stream",
+    rank = 2
+)]
+pub(super) async fn stream_logs<'r>(
+    app_name: Result<AppName, AppNameError>,
+    service_name: &'r str,
+    log_query: LogQuery,
+    apps: &'r State<Arc<Apps>>,
+) -> HttpResult<EventStream![Event + 'r]> {
+    let app_name = app_name.unwrap();
```
samuchila, line 67: To keep it uniform accross both functions - `let app_name = app_name?;`

api/src/apps/routes/logs.rs
```diff
@@ -0,0 +1,374 @@
...
+            ),
+            None => format!(
+                "/api/apps/{}/logs/{}?since={}",
+                self.app_name,
+                self.service_name,
+                RawStr::new(&from.to_rfc3339()).percent_encode(),
+            ),
+        };
+
+        let content_dispositon_value = if self.as_attachment {
```
samuchila, line 140: Self-Correction :) - My initial typo has carried through.
Please correct this - `content_disposition_value`

samuchila (approved):
Found a couple of my own minor mistakes. Please have a look at them.
Otherwise, the approach is great, approving the PR.

api/src/apps/routes/logs.rs
schrieveslaach: done, thanks for watching out.

Issue #146, opened 2023-11-30
Bootstrapping Companions


schrieveslaach:
The support of companions is very limited and is less capable of Docker Compose or Kubernetes files. For example, it is possible to specific files that will be put into the container but something like an additional job is impossible (some applications may rely on a oneshot database migration). This limitation restricts the use of PREvant to simple applications but applications composed of multiple complex applications are impossible to manage on PREvant.

Thus PREvant should be able to bootstrap the companions with a predefined set of configuration files (docker compose or Kubernetes files). Each infrastructure type can start an external container based on an OCI images that generates the desired file format. This format will be parsed and merged with the defined deployment unit before it will be deployed.

- [ ] Docker Compose bootstrapping
- [ ] Kubernetes bootstrapping

schrieveslaach:
On the architecture level PREvant's backends (Docker or Kubernetes) can start a set of containers that receive a configurable and templated list of arguments and then the containers pipe some Docker Compose or Kubernetes manifest into STDOUT. This can then be used by PREvant to have an set of companions and other configurations (such as volume mappings, see #123) in place. So, applications that can be expressed on Docker Compose or Kubernetes manifest could be deployment and enriched by PREvant.

The payload that PREvant receives in the request body will be merged into the parsed outputs and thus users can reconfigure certain values on the API level.

PR #163, opened 2024-03-12
Fix Duplicate Deployments in Bootstrapped Apps

schrieveslaach:
If the application is bootstrapped by a bootstap container and then companions have been converted into instances, and then the instances should be updated with a subsequent HTTP request, the application ends up with a second version of the instance. This circumstance is not intended and this commit makes sure that subsequent HTTP requests update existing deployments while taking the bootstrapping into account.

Related to #146

### DIFF
--- a/api/src/infrastructure/kubernetes/infrastructure.rs
+++ b/api/src/infrastructure/kubernetes/infrastructure.rs
@@ -64,7 +64,7 @@ use kube::{
 use log::{debug, warn};
 use multimap::MultiMap;
 use secstr::SecUtf8;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::convert::{From, TryFrom};
 use std::net::IpAddr;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -522,8 +522,22 @@ impl Infrastructure for KubernetesInfrastructure {
...
         let services = self.get_services_of_app(app_name).await?;
-        k8s_deployment_unit.filter_by_instances_and_replicas(&services);
+
+        k8s_deployment_unit.filter_by_instances_and_replicas(
+            services
+                .iter()
+                // we must not exclude the services that are provided by the deployment_unit
+                // because without that filter a second update of the service would create an
+                // additional deployment instead of updating/merging the existing one.
+                .filter(|s| !deployment_unit_service_names.contains(s.service_name())),
```
samuchila, line 539: I somehow do not follow this. This filter here effectively achieves exclusion at the end right?
Did you perhaps mean `we must exclude`?

samuchila: Yes, this looks right.

samuchila (approved):

PR #167, opened 2024-05-07
Migrate from Vue CLI to Vite

schrieveslaach:
Vue CLI is in maintenance mode and the building of the frontend fails in none root enviroments (postcss bug in cosmiconfig, see https://github.com/webpack-contrib/postcss-loader/issues/662#issuecomment-1870337757). This commit provides the migration to vite as the sucessor tool.


frontend/package.json
```diff
@@ -1,46 +1,34 @@
...
+      "@fortawesome/free-solid-svg-icons": "^5.15.4",
+      "@fortawesome/vue-fontawesome": "^3.0.3",
+      "@vue/compat": "^3.4.27",
+      "bootstrap-material-design": "^4.1.3",
+      "core-js": "^3.27.1",
+      "jquery": "^3.7.1",
+      "moment": "^2.30.1",
+      "http-link-header": "^1.1.3",
+      "popper.js": "^1.16.1",
+      "swagger-ui": "^4.19.0",
```
samuchila, line 22: Core-js has a new version - https://www.npmjs.com/package/core-js/v/3.37.1
Swagger-ui as well has newer versions which have more downloads.

frontend/vite.config.mjs
```diff
@@ -0,0 +1,44 @@
...
+   },
+   transpileDependencies: ["bootstrap-material-design"],
+   build: {
+      commonjsOptions: { transformMixedEsModules: true },
+   },
+   plugins: [
+      vue({
+         template: {
+            compilerOptions: {
+               // PREvant's UI is ported to Vue3 but the vue-virtual-scroller isn't.
```
samuchila, line 18: https://github.com/Akryum/vue-virtual-scroller/tree/next

I thought we use [v2.0.0-beta.8](https://github.com/Akryum/vue-virtual-scroller/releases/tag/v2.0.0-beta.8)
And this is already ported to vue 3.

samuchila (approved):
The changes look good, hence approving it. 
Please address the 2 very minor points raised.

frontend/vite.config.mjs
schrieveslaach: Thanks. I think I was confused by other stuff going on. I could get rid of some other workarounds in this file too.

frontend/package.json
schrieveslaach: core-js is update. Swagger not yet because there is a breaking change for PREvant that prevents the upgrade to 5.x. I want to do that another time.

PR #173, opened 2024-06-26
Usage section Update- Why different container environments? README.md

AnumDastgir:
In the Usage section, I have added as to why multiple container environments like Docker or Kubernetes should be used in PREvant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ of them.
 # Usage
 
 In this section, it is discussed how different container environments (Docker
-and Kubernetes) benefit PREvant.PREvant utilizes both Docker and Kubernetes
+and Kubernetes) benefit PREvant. PREvant utilizes both Docker and Kubernetes
 because they are exceptional tools for managing microservices but both
 environments function differently. Docker uses network connection and alias to
 communicate with each other and usually a load balancer like Traefik can enhance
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,12 +50,36 @@ of them.
 
 # Usage
 
-Have a look at the examples directory. There you can find examples that deploy PREvant in different container environments:
+In this section, it is discussed how different container environments (Docker
+and Kubernetes) benefit PREvant.PREvant utilizes both Docker and Kubernetes
```
schrieveslaach, line 54:
```suggestion
and Kubernetes) benefit PREvant. PREvant utilizes both Docker and Kubernetes
```

schrieveslaach (changes requested):
Thanks for providing this PR. But I'm not sure if I get the story right because the new paragraphs in the Section Usage are for me unrelated to the question how one can use PREvant.

I also think that the paragraphs provide confusing statements. For example,

- Traefik is not a load balancer, it is more and defines itself as “an [open-source](https://github.com/traefik/traefik) Edge Router”, see [Traefik's doc](https://doc.traefik.io/traefik/).
- I think you can do integrative testing with Docker 

I think we need to discuss that face 2 face a bit.

PR #174, opened 2024-06-26
Development section Update README.md

AnumDastgir:
Development section needed more structure and some punctuation marks were missing. Grammatical mistakes are intended to be  corrected.


schrieveslaach (approved):
Looks good to me. Can you run something like `gq{motion}` on the changed lines before merging?

PR #177, opened 2024-07-05
Allow Adjustment of Companion Routing

schrieveslaach:
This commit enables that the routing of companions can be adjusted. For example the routing configuration of Adminer needs additional middleware configs (X-Forwarded-Prefix) to be usable behind an URL with sub path.


mateusz-k-wojcik (approved):
For me code looks fine. Great Job! I approve the MR.

PR #187, opened 2024-07-24
Container Options Section Update README.md

AnumDastgir:
Elaborated a little in this section

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -35,7 +35,7 @@ storageClass = 'local-path'
 
 ## Container Options
 
-Create a table `containers` with following options:
+A table containers can be created with the with the option of Restricting the memory usage of containers, as shown below. You can create a table `containers` with following options:
 
 ```toml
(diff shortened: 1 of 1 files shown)


api/README.md
```diff
@@ -35,7 +35,7 @@ storageClass = 'local-path'
 
 ## Container Options
 
-Create a table `containers` with following options:
+A table containers can be created with the with the option of Restricting the memory usage of containers, as shown below. You can create a table `containers` with following options:
```
schrieveslaach, line 38:
```suggestion
The following table `containers` can be used to set some global options for all the OCI containers that PREvant deploys.
For example, restricting the memory usage of containers, as shown below. For full reference see the example below:
```

schrieveslaach (changes requested):
Thanks for aiming to improve the docs. I stumbled across the formulation and suggested a different wording. Please, check it out.

PR #204, opened 2024-09-03
Limit Number of Apps

### LINKED ISSUE #200: Limit the Number of Apps
While there is still no way of deleting unused apps automatically (see #149), sometimes people forget to clean up there testing app and thus they occupy the resources of the backend (K8s or Docker). When there are too many applications running on any backend, people will observe weird behaviours due to swapping, container restarts, etc. 

Therefore, a simple way of pretending these issues is to limit the number of applications being deployed by PREvant. Thus PREvant should offer an optional configuration option that allows people to limit the number of application being deployed.

[Here](https://github.com/aixigo/PREvant/blob/c06ffbd97bb5fd3715d1c29d7090d21ad235e9eb/api/src/apps/mod.rs#L255) should be the place to evaluate the option.

The configuration needs to take into account that there are different options (see #202) and should be implemented in such a way that you either use the limiting number or something else.

schrieveslaach:
This commit provides the new config option `applications.max` to limit the number of applications that can be deployed in total.

In addition to this PREvant is now able to crawl web host meta information via port forwarding which improves local developmont.

Fixes #200

### DIFF
--- a/api/src/apps/mod.rs
+++ b/api/src/apps/mod.rs
@@ -139,6 +139,10 @@ impl AppsService {
         })
     }
 
+    pub fn infrastructure(&self) -> &dyn Infrastructure {
+        self.infrastructure.as_ref()
+    }
+
     /// Analyzes running containers and returns a map of `app-name` with the
     /// corresponding list of `Service`s.
     pub async fn get_apps(&self) -> Result<MultiMap<AppName, Service>, AppsServiceError> {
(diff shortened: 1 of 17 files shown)


api/src/apps/mod.rs
```diff
@@ -252,6 +256,14 @@ impl AppsService {
         replicate_from: Option<AppName>,
         service_configs: &[ServiceConfig],
     ) -> Result<Vec<Service>, AppsServiceError> {
+        if let Some(app_limit) = self.config.app_limit() {
+            let apps = self.get_apps().await?;
+
+            if apps.keys().len() + 1 > app_limit {
```
schrieveslaach, line 262: Note to myself: this prevents updates of existing apps.

schrieveslaach: Fixed.

PR #205, opened 2024-09-03
PREvant in the Nutshell section Update README.md

AnumDastgir:
This section can be further improved and made more meaningful by adding information at the beginning that a detailed explanation can be found in a research paper and the link is provided in the 'Further Reading' section.

PR #206, opened 2024-09-04
Additions in Basic Terminology section

AnumDastgir:
I have added the discussed the visualization which explains the process of composing microservices with PREvant into reviewable Applications. Additionally, have added the explanation for this visualization, which is inspired from the research paper explaining details.


README.md
```diff
@@ -50,6 +50,20 @@ manages the following kind of services:
...
+for microservices (Ms) Ms A, Ms B, and Ms C. During the build stage, each
+microservice is packaged as a container image and then pushed to a container
+image registry (e.g., a Docker registry), ensuring the services are prepared for
+deployment in the acceptance and manual test stages. In the deployment phase,
+such as the manual test stage shown in above Figure, the continuous delivery
+pipeline can leverage PREvant’s REST API. This REST request creates a
+software-defined network, starts the microservice container, connects it to the
+network, and configures a reverse-proxy, making the service accessible via
+PREvant’s web interface. Subsequent REST calls check for newer versions of the
+container image, and if one is found, the container is updated.
```
schrieveslaach, line 65: That paragraph is a quote from the paper and we should mark it as such.

```suggestion
> Figure above depicts the separate repositories and continuous delivery pipelines
> for microservices (Ms) Ms A, Ms B, and Ms C. During the build stage, each
> microservice is packaged as a container image and then pushed to a container
> image registry (e.g., a Docker registry), ensuring the services are prepared for
> deployment in the acceptance and manual test stages. In the deployment phase,
> such as the manual test stage shown in above Figure, the continuous delivery
> pipeline can leverage PREvant’s REST API. This REST request creates a
> software-defined network, starts the microservice container, connects it to the
> network, and configures a reverse-proxy, making the service accessible via
> PREvant’s web interface. Subsequent REST calls check for newer versions of the
> container image, and if one is found, the container is updated. [(Paper titlle)](link to paper)
```

Also, please mark the changes you made to words with square brackets, like described [here](https://writingcommons.org/article/inserting-or-altering-words-in-a-direct-quotation/).

README.md
```diff
@@ -50,6 +50,20 @@ manages the following kind of services:
   Alternatively, any other application can be specified as a source of
   replication.
 
+![Composing Microservices](assets/composing-microservices.png "Composing Microservices")
```
schrieveslaach, line 53: We should keep the original image from the paper because it matches to the screenshot above.

Please, also try to use it as SVG because the other images are also in SVG and then with to the page width.

schrieveslaach (changes requested):

README.md
schrieveslaach, line 53: I'm wondering if this is the right spot. I think this image and paragraph should be above the first new paragraph that is introduce by #205 (in the nutshell section). with an additional sentence above the image, it should provide a good reading flow.

schrieveslaach: Or it should go above the explanation of instance and replica to tell someone what how a composition of services should look like.

[Scratch my comment above](https://github.com/aixigo/PREvant/pull/206/files#r1743320841). I like this variant better.

AnumDastgir: Done

PR #212, opened 2024-11-04
Improve templating with user defined parameters

schrieveslaach:
This PR brings user defined templating to companions in case of static configuration and bootstrapped from the infrastructure. This change provides the ability to implement more dynamic deployment scenarios.

### DIFF
--- a/api/res/openapi.yml
+++ b/api/res/openapi.yml
@@ -83,9 +83,21 @@ paths:
         content:
           application/json:
             schema:
-              type: array
-              items:
-                $ref: '#/components/schemas/ServiceConfiguration'
+              oneOf:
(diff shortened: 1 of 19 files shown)


api/res/openapi.yml
```diff
@@ -83,9 +83,21 @@ paths:
...
+                - type: array
+                  items:
+                    $ref: '#/components/schemas/ServiceConfiguration'
+                - type: object
+                  properties:
+                    services:
+                      type: array
+                      items:
+                        $ref: '#/components/schemas/ServiceConfiguration'
+                    userDefined:
```
schrieveslaach, line 96: Not sure yet about this name.

PR #245, opened 2025-05-20
Fix resetting of filters when closing OpenAPI documentation

### LINKED ISSUE #244: Closing OpenAPI Documentation via 'X' resets the application name filter
# Steps to reproduce:
1. Search for an application using its name.
2. Open the OpenAPI documentation for any service.
3. Close the popup window with the `X` button on the top right.
4. You are rerouted to `/#`, losing your previous filter and showing all applications.

Note: Using the "back" button of your browser brings you back to the homepage, with all your filters intact.

# Analysis
The `close()` method in `OpenApiUI.vue` uses `.push('/')` to send you back to the homepage. Replacing this call with `.back()` would provide the same functionality as using the browsers 'back' button and fixing this bug.

MLesker:
Closes #244

### DIFF
--- a/frontend/src/OpenApiUI.vue
+++ b/frontend/src/OpenApiUI.vue
@@ -135,7 +135,7 @@
       },
       methods: {
          close() {
-            this.$router.push('/');
+            this.$router.back();
          }
       }
    }


schrieveslaach:
Thanks for the contribution. Do you mind to check if the other dialogs also have the same issue?

MLesker:
Thanks, I actually found two other occurrences and fixed them too.

PR #268, opened 2025-11-11
feat(frontend): allow configuring default app name

drik98:
In the past the default app name was considered to be "master". With this change it will be possible to configure the name of the default app while remaining backwards compatible.

### DIFF
--- a/.github/workflows/vitest-unit-tests.yml
+++ b/.github/workflows/vitest-unit-tests.yml
@@ -31,10 +31,3 @@ jobs:
 
       - name: Run Vitest Unit Tests
         run: npm run test:unit
-
-      - name: Upload Coverage Report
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -92,6 +92,15 @@ PREvant provides configuration options to customize the frontend. Currently, you
 title = 'My Environment'
 ```
 
+and the name of the default app: 
+
+```toml
+[frontend]
+default_app_name = 'latest'
```
schrieveslaach: **note:** As said above should be part of the `applications` section.

schrieveslaach (approved):
Very nice, thanks for addressing my remarks. I'll merge it into #271 and then merge both commits into main. I'll also do a short smoke test before.

PR #273, opened 2025-11-22
feat(frontend): improve ux for required auth

drik98:
- when PREvant is configured to require auth then the user needs to be logged in in order to perform writing operations such as creating tiles or deleting ones
- PREvant currently offers to manually delete and duplicate apps directly through the UI
- before the user could trigger the operation but the REST-Api rejected the call correctly but left the user confused as to why it was not possible
- with this change the duplication/deletion dialogs can still be opened but the actual inputs to perform the action are disabled and a warning alert is displayed  
  
image

- the dialogs can still be opened and the menu options are not hidden/disabled so we are able to give the user better feedback as to why it cannot be done and what he has to do
- also fixed some vue warnings because the annoyed me
- also fixed the regression of not being able to start and stop services
- we can also move the former two points to separate PRs

### DIFF
--- a/frontend/src/scss/_basics.scss
+++ b/frontend/src/scss/_basics.scss
@@ -41,6 +41,10 @@ a {
    pointer-events: none;
 }
 
+input.form-control {
+   padding: .28125rem !important;
+}
+
 .ra-ellipsis {
    overflow: hidden;
 


drik98:
Do we also want something like

image

?

frontend/src/components/DuplicateAppDialog.vue
```diff
@@ -32,14 +32,19 @@
...
          </div>
+         <div v-if="!hasWritePermissions" class="alert alert-warning text-center" role="alert">
+            You need to be logged in to duplicate apps.
+         </div>
       </template>
       <template v-slot:footer>
          <button
                type="button"
                class="btn btn-outline-primary"
+               :disabled="!hasWritePermissions || newAppName.length === 0"
```
schrieveslaach: **question:** When is `newAppName.length === 0` the case?

schrieveslaach (approved):
Thank you very much for improving the UX, I really appreciate it.

There is just one question I have, nothing big that prevents merging.

PR #289, opened 2026-02-06
Metric Based Automatic Clean-up

### LINKED ISSUE #149: Automatic Cleanup for PREvant tiles
PREvant allows us to instantiate individual environments very quickly. While this being a great feature, these easily created environments can easily be forgotten, thus causing an issue of draining and running out of resources at the hosting system.

To overcome this issue, an automatic cleanup for those unwanted environments is needed, so that if they're not being used, they should release the occupied resources automatically. Therefore, we consider following implementation plan:

1. PREvant should gather OpenTelemetry metrics from Traefik (more specific, [HTTP metrics](https://doc.traefik.io/traefik/reference/install-configuration/observability/metrics/#http-metrics)) to check if applications' routes haven’t been accessed in recent. 
2. The least used applications will then be backed-up in database, e.g. the Kubernetes YAML manifests of application, and the applications will be deleted the infrastructure, except persistent volumes
3. The application can be restored from the database and deployed with the preserved persistent volumes
4. If the applications haven't restored from database in a period of time, the database and existing persistent volumes will be cleaned up

Implementation tasks:

- [x] manual back-up & restore of applications, see #281
- [ ] back-up & restore from frontend
- [ ] detect if applications are stale
- [ ] automatic back-up stale applications

schrieveslaach:
Fixes #149

### DIFF
--- a/api/src/apps/clean_up.rs
+++ b/api/src/apps/clean_up.rs
@@ -1,0 +1,681 @@
+use crate::{
+    apps::{repository::AppPostgresRepository, AppTaskQueueProducer, Apps},
+    config::{ApplicationCleanUpPolicy, Config},
+    models::AppName,
+};
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use futures::try_join;
+use regex::Regex;
+use rocket::{
+    fairing::{Fairing, Info, Kind},
+    Build, Orbit, Rocket,
+};
+use std::{
+    collections::{BTreeMap, HashMap},
+    sync::atomic::AtomicBool,
+    time::Duration,
+};
+
+pub struct AppCleanUp {}
+
+impl AppCleanUp {
+    pub fn fairing() -> Self {
+        Self {}
+    }
+}
+
+#[rocket::async_trait]
+impl Fairing for AppCleanUp {
+    fn info(&self) -> Info {
+        Info {
+            name: "app-clean-up",
+            kind: Kind::Ignite | Kind::Liftoff,
+        }
+    }
+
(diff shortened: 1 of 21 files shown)


api/src/apps/clean_up.rs
```diff
@@ -0,0 +1,556 @@
...
+            log::debug!("Checking for stale applications");
+
+            let app_names_to_be_considered_for_back_up = {
+                let apps = app_updates.borrow().clone();
+
+                let time_to_exists_before = Utc::now() - self.clean_up_policy.time_to_use;
+                apps.into_iter()
+                    .filter(|(_, app)| {
+                        app.created_at
+                            .is_none_or(|created_at| created_at <= time_to_exists_before)
```
schrieveslaach: **issue:** @aixigo-dvoss reported that just after creation the tile has been immediately backed up. Note to myself: we need some unit testing here.

docs/configuration.md
```diff
@@ -79,6 +79,77 @@ defaultApp = "master"
...
+# applications won't be moved to backups.
+#
+# (optional): By default this will be equal to the defaultApp.
+permanentApps = [ "master" ]
+
+# This configuration option tells PREvant when there are the busy hours with a
+# UTC based cron expression. When this option is set, PREvant will not perform
+# any automated backups.
+# (optional): not set by default
+busyHours = { start = "0 8 * * * Mon-Fri", end = "0 16 * * * Mon-Fri" }
```
schrieveslaach: **todo:** Improve the example here because the busy hours would be every hour from 8 minutes after to 18 minutes after. Which is weird.

```suggestion
busyHours = { start = "0 0 8 * * Mon-Fri", end = "0 0 16 * * Mon-Fri" }
```

PR #303, opened 2026-02-25
Better Navigation Among Applications

schrieveslaach:
With this change a user of the dashboard can quickly navigate among the sections. For example, they can jump to the backups or to their previews via the footer. Additionally, the footer shows how many applications are present in each section.

### DIFF
--- a/frontend/src/Main.vue
+++ b/frontend/src/Main.vue
@@ -3,8 +3,10 @@
    <main>
       <RouterView />
    </main>
+   <Footer/>
 </template>
 
 <script setup>
+import Footer from './components/Footer.vue'
 import Navbar from './components/Navbar.vue'
 </script>
--- a/frontend/src/components/Footer.vue
+++ b/frontend/src/components/Footer.vue
@@ -1,0 +1,56 @@
+<template>
+   <footer class="footer fixed-bottom text-light bg-dark" v-if="isAnySectionAvailable">
+      <div class="container-fluid text-center text-md-left">
+         <div class="row">
+            <div class="col-md-12">
+               <div class="text-center py-3">
+                  <RouterLink :to="getLinkTo('my-previews')" v-if="ownedApps.length > 0">
+                     My Previews ({{ ownedApps.length }})
+                  </RouterLink>
+                  <RouterLink :to="getLinkTo('apps-without-tickets')" v-if="appsWithoutTicket.length > 0">
+                     Previews ({{ appsWithoutTicket.length }})
+                  </RouterLink>
+                  <RouterLink :to="getLinkTo('apps-with-tickets')" v-if="appsWithTicket.length > 0">
+                       Features ({{ appsWithTicket.length }})
+                  </RouterLink>
+                  <RouterLink :to="getLinkTo('backed-up-apps')" v-if="appBackups.length > 0">
+                       Backups ({{ appBackups.length }})
+                  </RouterLink>
+                  <span v-if="appNameFilter != ''">
+                     <font-awesome-icon icon="exclamation"/> Filtered by <em>{{ appNameFilter }}</em>
+                  </span>
+               </div>
+            </div>
+         </div>
+      </div>
+   </footer>
+</template>
+
+<style scoped>
+footer a {
+   padding: 0.5em;
+}
+</style>
+
+<script setup>
+   import { computed } from 'vue';
+   import { useStore } from 'vuex';
+   import { useRoute } from "vue-router";
+
+   const store = useStore();
+   const route = useRoute();
+
+   const appsWithTicket = computed(() => store.getters.appsWithTicket);
+   const appsWithoutTicket = computed(() => store.getters.appsWithoutTicket);
+   const appBackups = computed(() => store.getters.appBackups);
+   const ownedApps = computed(() => store.getters.ownedApps);
+   const appNameFilter = computed(() => store.getters.appNameFilter);
+
+   const isAnySectionAvailable = computed(() => {
+      return store.getters.ownedApps.length > 0 || store.getters.appsWithTicket.length > 0 || store.getters.appsWithoutTicket.length > 0 || store.getters.appBackups.length > 0;
+   })
+
+   function getLinkTo(heading) {
+      return { query: route.query, params: { heading } };
+   }
+</script>
--- a/frontend/src/main.js
+++ b/frontend/src/main.js
@@ -44,7 +44,7 @@ import OpenApiUI from './views/OpenApiUI.vue';
 import LogsDialog from './views/LogsDialog.vue';
 
 import {library} from '@fortawesome/fontawesome-svg-core';
-import {faClipboard, faCode, faCopy, faServer, faSpinner, faTerminal, faTrash, faWindowClose, faDownload} from '@fortawesome/free-solid-svg-icons';
+import {faClipboard, faCode, faCopy, faServer, faSpinner, faTerminal, faTrash, faWindowClose, faDownload, faExclamation} from '@fortawesome/free-solid-svg-icons';
 import {FontAwesomeIcon} from '@fortawesome/vue-fontawesome';
 import { createStore } from './store';
 
@@ -57,12 +57,13 @@ library.add(faSpinner);
 library.add(faTerminal);
 library.add(faTrash);
 library.add(faWindowClose);
+library.add(faExclamation);
 
 export const router = createRouter({
    history: createWebHashHistory(),
    // It is currently not possible to use lazy loading for routes because of bootstrap v4 and jquery
    routes: [
-      { path: '/', component: Apps, query: { appNameFilter: { type: String } } },
+      { path: '/:heading?', component: Apps, query: { appNameFilter: { type: String } } },
       { path: '/open-api-ui/:url', name: 'open-api-ui', component: OpenApiUI },
       { path: '/async-api-ui/:url', name: 'async-api-ui', component: AsyncApiUI },
       { path: '/logs/:app/:service', name: 'logs', component: LogsDialog }
--- a/frontend/src/views/Apps.vue
+++ b/frontend/src/views/Apps.vue
@@ -38,7 +38,7 @@
          No apps to review.
       </h1>
 
-      <h1 class="ra-container__title--preview" v-if="ownedApps.length > 0">My Previews</h1>
+      <h1 id="my-previews" class="ra-container__title ra-container__title--preview" v-if="ownedApps.length > 0">My Previews</h1>
       <transition-group tag="div" name="list-complete" class="ra-container__apps--preview ra-apps ">
          <review-app-card
             v-for="reviewApp in ownedApps"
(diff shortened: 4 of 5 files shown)


frontend/src/components/Footer.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+   <footer class="footer fixed-bottom text-secondary bg-dark" v-if="showFooter">
```
MLesker: **suggestion:** Visibility of this bar (especially the text) is IMO currently suboptimal. Maybe it's due to the bright white right above it, but I find it hard to read the titles. Maybe adjust the contrast?

Image

frontend/src/components/Footer.vue
```diff
@@ -0,0 +1,65 @@
...
+
+<style scoped>
+footer a {
+   padding: 0.5em;
+}
+</style>
+
+<script>
+import { mapGetters } from 'vuex';
+export default {
```
drik98: **suggestion:** I would prefer if completely new components would be using the current vue idiomatic `script setup` components instead of the deprecated options API.

frontend/src/components/Footer.vue
```diff
@@ -0,0 +1,65 @@
...
+         }
+         if (this.appsWithoutTicket.length > 0) {
+            ++count;
+         }
+         if (this.appBackups.length > 0) {
+            ++count;
+         }
+
+         return count > 1;
+      }
```
drik98: **nitpick:** This seems unnecessarily verbose and complicated, with unnecessary executed branches. 

**suggestion:** Reflect the same conditions independently used in the template section.

 
```suggestion
      showFooter() {
         return this.ownedApps.length > 0 || this.appsWithTicket.length > 0 || this.appsWithoutTicket.length > 0 || this.appBackups.length > 0
      }
```

frontend/src/Main.vue
```diff
@@ -3,8 +3,10 @@
    <main>
       <RouterView />
    </main>
+   <Footer/>
```
drik98, line 6: **issue:** (Possibly) we are mounting the footer in a global context, but the links within this footer are context-dependend to the `Apps` component. This should at the moment not really be an issue as the other routes are all fake dialogs so one cannot click the link anyways. 

An improvement would be to to also add the name/path in the `getLinkTo` method, just to be safe. (That would require to add a `name` to the `Apps` route definition)

frontend/src/views/Apps.vue
```diff
@@ -138,6 +150,31 @@
          changeServiceState( appName, serviceName ) {
             this.$store.dispatch( 'changeServiceState', { appName, serviceName } );
          }
+      },
+      watch: {
+         reviewApps: {
```
drik98, line 158: **issue:** Due to the watcher on reviewApps, the auto scroll is triggered repeatedly and cause jumpy UX. 

Example: Click on "Features" and you are navigated to the features section. Scroll somewhere else. Then an update from the apps eventstream comes, which triggers the watcher again and you are taken to the "Features" section again and again.

frontend/src/views/Apps.vue
```diff
@@ -138,6 +150,31 @@
...
+            },
+         },
+         $route (to) {
+            const heading = to.params.heading;
+            if (heading) {
+               const element = document.getElementById(heading);
+               if (element) {
+                  element.scrollIntoView({ behavior: 'smooth' });
+               }
+            }
```
drik98: **suggestion:** This logic is duplicated and could be extracted into a function.

frontend/src/main.js
```diff
@@ -62,7 +62,7 @@ export const router = createRouter({
    history: createWebHashHistory(),
    // It is currently not possible to use lazy loading for routes because of bootstrap v4 and jquery
    routes: [
-      { path: '/', component: Apps, query: { appNameFilter: { type: String } } },
+      { path: '/:heading?', component: Apps, query: { appNameFilter: { type: String } } },
```
drik98, line 66: **issue:** I would not use a dynamic path param at this level as we are already having other routes at the same level. 

Just an example but if we had an element with id "open-api-ui" then it would not work anymore. If we even need to modify the router i would suggest to go with a query param instead.

Though I don't think we even need to adjust the route path or query. When constructing the `to`-route in the Footer, we can pass a hash e.g. 

```
{
name: 'apps',
hash: `#my-previews`, // to reference the section
query:  { ...route.query } // dont loose the current query
}
```

in combination with the before mentioned `scrollBehavior` 

```
export const router = createRouter({
  history: createWebHashHistory(),
  scrollBehavior(to, from, savedPosition) {
    if (savedPosition) {
      return savedPosition;
    }

    if (to.hash) {
      return {
        el: to.hash,
        behavior: "smooth",
      };
    }

    return { top: 0 };
  },
});
```

should achieve the excpected result without any side effects.

frontend/src/components/Footer.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+   <footer class="footer fixed-bottom text-secondary bg-dark" v-if="showFooter">
+     <div class="text-center py-3">
+        <RouterLink :to="getLinkTo('my-previews')" v-if="ownedApps.length > 1">
+           Previews: {{ ownedApps.length }}
```
drik98: **todo:** You probably meant


```suggestion
           My Previews: {{ ownedApps.length }}
```

frontend/src/components/Footer.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+   <footer class="footer fixed-bottom text-secondary bg-dark" v-if="showFooter">
+     <div class="text-center py-3">
+        <RouterLink :to="getLinkTo('my-previews')" v-if="ownedApps.length > 1">
```
drik98: **issue:** This won't render the link when we only have 1 preview. We probably want


```suggestion
        <RouterLink :to="getLinkTo('my-previews')" v-if="ownedApps.length > 0">
```

or `>= 1`

drik98: Same for other instances.

frontend/src/views/Apps.vue
```diff
@@ -85,6 +85,18 @@
...
+   }
+   #apps-with-tickets {
+      scroll-margin-top: 100px;
+   }
+   #apps-without-tickets {
+      scroll-margin-top: 100px;
+   }
+   #backed-up-apps {
+      scroll-margin-top: 100px;
+   }
```
drik98: **issue:** Ids should only be used for styling as last resort. Instead use a common class to apply the styling to make it also more easily extendable. 

**suggestion:** Apply the class `ra-container__title` to every h1 element given that we already have for example `ra-container__title--feature` and `ra-container__title--preview` following BEM.

frontend/src/views/Apps.vue
```diff
@@ -138,6 +150,31 @@
...
+         },
+         $route (to) {
+            const heading = to.params.heading;
+            if (heading) {
+               const element = document.getElementById(heading);
+               if (element) {
+                  element.scrollIntoView({ behavior: 'smooth' });
+               }
+            }
+         }
```
drik98, line 172: **suggestion:** Consider using https://router.vuejs.org/guide/advanced/scroll-behavior API from vue-router instead of coming up with a custom solution.

frontend/src/components/Footer.vue
```diff
@@ -0,0 +1,65 @@
+<template>
```
drik98, line 1: **thought:** The displayed counts (if the link is displayed at all) are currently dependend on the filter of the app name. This could potentially be misleading to the user. 

We have to think about if we want to keep this behavior or and if yes, we we should display a hint that the count is currently differing from the total count due to applied filtering.

frontend/src/components/Footer.vue
schrieveslaach: Your code is not equivalent to the logic. I want to hide the footer if there is only one section (my reasoning: what is the point of having a link to the only place that is available that is already visible). However, the naming is not optimal. I'll improve.

frontend/src/views/Apps.vue
schrieveslaach: Thanks for the hint. I'll check it out.

frontend/src/components/Footer.vue
drik98: I would argue even with only one section it is valuable to see how many apps it contains.

drik98: If you do want to keep the behavior if would suggest adding a quick unit component test to document the intention. 

"should not render the footer when only one section exists"
"should not render the footer when no section exists"
"should render the footer when at least two sections exist"

frontend/src/views/Apps.vue
schrieveslaach: Done

frontend/src/views/Apps.vue
schrieveslaach: Done. I did not use the vue-router scroll behavior because I couldn't figure out to wait reliably for applications being available and when the applications are rendered into sections.

frontend/src/views/Apps.vue
schrieveslaach: done.

frontend/src/views/Apps.vue
schrieveslaach: see comment above.

frontend/src/main.js
schrieveslaach: The thing is that scroll to must wait until the headings are visible. Is there a way to wait for it without using timeout?

frontend/src/components/Footer.vue
schrieveslaach: I changed it to at least one section and I added the note:

Image

frontend/src/components/Footer.vue
schrieveslaach: Image

frontend/src/components/Footer.vue
schrieveslaach: Image

frontend/src/views/Apps.vue
drik98: Hmm, understood.

frontend/src/views/Apps.vue
```diff
@@ -137,6 +139,32 @@
       methods: {
          changeServiceState( appName, serviceName ) {
             this.$store.dispatch( 'changeServiceState', { appName, serviceName } );
+         },
+         scrollToSection(section) {
+            this.$nextTick(function () {
```
drik98: **suggestion:** Could also use async/await instead of callbacks, _I think._



```suggestion
        async scrollToSection(section) {
            await this.$nextTick();
            const element = document.getElementById(section);
            ...
```

Might need to check out if this is actually working in the options API.

drik98 (approved):

frontend/src/components/Footer.vue
schrieveslaach: Done.

frontend/src/views/Apps.vue
schrieveslaach: That works. I applied it.

frontend/src/views/Apps.vue
schrieveslaach: I left a comment explaining why I used watcher instead of scroll-behavior.

PR #308, opened 2026-02-27
Make Services Available in Companion Bootstrapping

### LINKED ISSUE #304: Make Services Available in Companion Bootstrapping
Currently, PREvant allows to use the list of services that will be deployed for an application when using the templating for the static companions, see [docs](https://github.com/aixigo/PREvant/blob/b942aa45d5e0f2a004faab25de2f3b254035f61d/docs/companions.md):

> `services`: An array of the services of the application. Each element has the following structure:
> - `name`: The service name which is equivalent to the network alias
> - `port`: The exposed port of the service
> - `type`: The type of service. For example, instance, replica, app-companion, or service-companion.

However, this templating variable is not available for bootstrapped companions, see [docs](https://github.com/aixigo/PREvant/blob/b942aa45d5e0f2a004faab25de2f3b254035f61d/docs/companions.md

> - `application`: The companion's application information
>   - `name`: The application name
>   - `baseUrl`: The URL that all services in the application share
> - `infrastructure`: Holds information that are specific to the underlying
>   infrastructure.
>   - 'namespace' (Kubernetes): the name of the namespace where the application
>     will be installed in.
> - `userDefined`: see [user defined schema configuration](user-defined-schema.md).

This list should be extended by:

> - `application`: The companion's application information
>   - `name`: The application name
>   - `baseUrl`: The URL that all services in the application share
> - `infrastructure`: Holds information that are specific to the underlying
>   infrastructure.
>   - 'namespace' (Kubernetes): the name of the namespace where the application
>     will be installed in.
> - `userDefined`: see [user defined schema configuration](user-defined-schema.md).
> - `services`: : An array of the services of the application. Each element has the following structure:
>   - `name`: The service name which is equivalent to the network alias.
>   - `image`: The service's image.
>   - `type`: The type of service. For example, instance, replica, app-companion, or service-companion.

With this additional template variable, one can design an additional bootstrap container argument

```toml
[[companions.bootstrapping.containers]]
image = "registry.example.com/user/bootstrap-helm-chart:latest"
args = [
   "--services", """/{{#each service}}
   {{~#isNotCompanion type~}}
   {{name}}@{{image}},
   {{~/isNotCompanion~}}
   {{/each}}"""
]
```

This config would be then translated to the command line arguments `--services`, `db@mariadb:latest,blog@wordpress@latest` if one uses the [the wordpress example from the OpenAPI spec](https://github.com/aixigo/PREvant/blob/b942aa45d5e0f2a004faab25de2f3b254035f61d/api/src/openapi.rs#L109-L164).

schrieveslaach:
Work in progress implementation

Fixes #304, depends on #317

### DIFF
--- a/api/src/infrastructure/kubernetes/infrastructure.rs
+++ b/api/src/infrastructure/kubernetes/infrastructure.rs
@@ -969,7 +969,7 @@ impl Infrastructure for KubernetesInfrastructure {
             let middleware = TraefikMiddleware::from_json(
                 middleware.metadata.name.expect("There should be a name"),
                 middleware.spec.0,
-            );
+            )?;
 
             traefik_middlewares.push(middleware);
         }
@@ -1046,18 +1046,16 @@ impl Infrastructure for KubernetesInfrastructure {
         &self,
         context: MergeRawElementsContext<'_>,
         raw_elements: Vec<RawInfrastructureElement>,
-    ) -> Vec<RawInfrastructureElement> {
-        K8sDeploymentUnit::parse_from_json(
+    ) -> Result<Vec<RawInfrastructureElement>> {
+        Ok(K8sDeploymentUnit::parse_from_json(
             &AppName::master(),
             raw_elements.into_iter().map(serde_json::Value::from),
-        )
-        // TODO: the interface must have error handling
-        .unwrap()
-        .update_with_merge_context(context)
+        )?
+        .update_with_merge_context(context)?
         .to_json_vec()
         .into_iter()
         .map(RawInfrastructureElement::from)
-        .collect::<Vec<_>>()
+        .collect::<Vec<_>>())
     }
 
     async fn resolve_infrastructure_template_data(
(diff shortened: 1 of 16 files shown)


api/src/infrastructure/kubernetes/infrastructure.rs
```diff
@@ -657,14 +657,19 @@ impl Infrastructure for KubernetesInfrastructure {
 
         let client = self.client().await?;
 
-        let app_name = deployment_unit.app_name();
+        let app_name = &deployment_unit.app_name;
         let bootstrapping_containers = self.config.companion_bootstrapping_containers(
```
schrieveslaach: **thought:** It feels like that the templating logic must move elsewhere and must be unified with the templating for static companions. For example, the `configs_for_templating` could be used for this purpose. I think I need to invest here and make a domain module that holds such logic independent of the infrastructure (like in the Hamburger-Architecture, see https://www.youtube.com/watch?v=RJ0k6_MgsN4)

schrieveslaach: Done with #317

PR #329, opened 2026-06-11
Idempotent Generation of Random Generated Secrets in Companion Bootstrapping

### LINKED ISSUE #328: Idempotent Generation of Random Generated Secrets in Companion Bootstrapping
There are case where it is necessary to generate random passwords for an application, e.g. an OpenID client secret that is shared between a Keycloak instance and an API gateway, during the bootstrapping of applications in Kubernetes environments. However, currently the bootstrap container do not have a change to see if secrets have been generated or not. Therefore, PREvant must inform the bootstrap container that a previous deployment for an application already created a set of secrets. 

This can be achieved by labeling the secret from which bootstrap container they originate and in a second run these secrets will be mounted into the bootstrap container for lookups.

schrieveslaach:
Fixes #328

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,7 +20,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


schrieveslaach:
This PR hasn't brought the fix for idempotent secret generation. It was opened with c8327a1 and while i was working on bc951ce4011d58561ca3b10b77714c0f5913692e I decided to push the c8327a1 to mainline branch. This then marked the PR as completed.

